PR #5, opened 2019-05-21
fixbug : parse private versioning yarn.lock

tomoyamachi:
fix parse bug on yarn.lock.
ie) `"jquery@git+https://xxxx:x-oauth-basic@github.com/tomoyamachi/jquery":`

https://github.com/knqyf263/trivy/issues/39#issuecomment-494365349


knqyf263:
👍

PR #7, opened 2019-07-09
Fix cargo's test on *nix link OS

iwamatsu:
Unix link OS distinguish between filename case. The test fails because the
filename used in the test are different in case. This changes the case of
the filename to lower case.


knqyf263:
I'm so sorry, but I didn't notice your PR... I've fixed this problem [here](https://github.com/aquasecurity/go-dep-parser/pull/11)

Let me close this PR. Thank you for your awesome contribution.

PR #12, opened 2020-03-29
add parser for go modules

yashbhutwala:


pkg/gomod/parse_testcase.go
```diff
@@ -0,0 +1,19 @@
+package gomod
+
+import "github.com/aquasecurity/go-dep-parser/pkg/types"
+
+var (
+	// docker run --name bundler --rm -it ruby:2.6 bash
+	// bundle init
+	// bundle add dotenv json faker rubocop pry
+	// bundler show | grep "*" | grep -v bundler | awk '{if(match($0, /\((.*)\)/)) printf("{\""$2"\", \""substr($0, RSTART+1, RLENGTH-2)"\"},\n");}'
```
knqyf263, line 9: This comment should illustrate how we generate these golden data.
For example,

```
docker run --name gomod --rm -it golang:1.14 bash 
cd src
go mod init example.com/m
go get github.com/aquasecurity/go-dep-parser
go list -m all | awk '{printf("{\""$1"\", \""$2"\"},\n")}'
```

yashbhutwala:
@knqyf263 sorry it's been a while 😄, thanks for taking a look and giving some feedback!  afaik, go.sum is not a lock file: https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using  and even though it's recommended to commit go.sum file with go.mod file, it is not necessary for someone to push it with their code: https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file

I could get indirect dependencies via https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod and https://pkg.go.dev/golang.org/x/mod/modfile?tab=doc#Require

Ideally, I could do something like `go list -m -json all`, lemme see.

knqyf263:
Oh, I didn't know `go.sum` is not a lock file. Thank you for telling me! However, as the document says, indirect dependencies are added `ff a dependency of your module does not itself have a go.mod (e.g., because the dependency has not yet opted in to modules itself), or if its go.mod file is missing one or more of its dependencies (e.g., because the module author did not run go mod tidy)`. It looks like the indirect dependencies are usually not included in `go.mod`.

As you can see in this example, `go.mod` contains only 1 package.
https://blog.golang.org/using-go-modules

```
$ cat go.mod
module example.com/hello

go 1.12

require rsc.io/quote v1.5.2
```

But, `go list -m all` shows more dependencies.

```
$ go list -m all
example.com/hello
golang.org/x/text v0.0.0-20170915032832-14c0d48ead0c
rsc.io/quote v1.5.2
rsc.io/sampler v1.3.0
```

>Another reason you might have // indirect dependencies in your go.mod file is if you have upgraded (or downgraded) one of your indirect dependencies beyond what is required by your direct dependencies, such as if you ran go get -u or go get foo@1.2.3. The go tooling needs a place to record those new versions, and it does so in your go.mod file (and it does not reach down into your dependencies to modify their go.mod files).

If we run `go get golang.org/x/text`, it is supposed to be contained with `indirect` comment in `go.mod`.

```
$ cat go.mod
module example.com/hello

go 1.12

require (
    golang.org/x/text v0.3.0 // indirect
    rsc.io/quote v1.5.2
)
```

If we don't run it, we can't see those indirect dependencies in `go.mod`, right? In short, I think `go.mod` is usually different from `go list -m all` we really need. If I overlook something, let me know, please.

knqyf263:
Close in favor of https://github.com/aquasecurity/go-dep-parser/pull/21

PR #13, opened 2020-09-03
WIP: add yarn v2 lockfile compatibility 

antoinedeschenes:
Yarn 2 uses a new lockfile format, which is now valid yaml. (Causing issue https://github.com/aquasecurity/trivy/issues/625 in Trivy)

This PR:
- Skips the __metadata top-level object which would otherwise trigger a "not package format" error:
```
__metadata:
  version: 4
  cacheKey: 6
```
- Parses version with various indents, with/without colon and double-quotes
```
  version: 1.2.3
--
  version: "1.2.3"
--
    version     1.2.3
```
- Parses v2 format "protocol" to skip non-npm package:
```
"code@workspace:.":
  version: "0.0.0-use.local"
--
"package@virtual:...":
```

Yarn lockfile examples:
Yarn v1:
```
# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
# yarn lockfile v1


"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4":
  version "7.10.4"
  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.10.4.tgz#168da1a36e90da68ae8d49c0f1b48c7c6249213a"
  integrity sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==
  dependencies:
    "@babel/highlight" "^7.10.4"
```

Yarn v2:
```
# This file is generated by running "yarn install" inside your project.
# Manual changes might be lost - proceed with caution!

__metadata:
  version: 4
  cacheKey: 6

"@babel/code-frame@npm:^7.0.0, @babel/code-frame@npm:^7.10.4":
  version: 7.10.4
  resolution: "@babel/code-frame@npm:7.10.4"
  dependencies:
    "@babel/highlight": ^7.10.4
  checksum: 05245d3b22a3ae849439195c4ee9ce9903dfd8c3fcb5124e77923c45e9f1ceac971cce4c61505974f411a9db432949531abe10ddee92937a0a9c306dc380a5b2
  languageName: node
  linkType: hard
```


pkg/yarn/parse_testcase.go
```diff
@@ -2316,4 +2317,523 @@ var (
 		{"yargs-parser", "9.0.2"},
 		{"yeast", "0.1.2"},
 	}
+
+	// docker run --name composer --rm -it -w /code node:12-alpine sh
```
knqyf263, line 2321:
```suggestion
	// docker run --name yarn2 --rm -it -w /code node:12-alpine sh
```

pkg/yarn/parse_testcase.go
```diff
@@ -2316,4 +2317,523 @@ var (
...
+	// yarn set version berry
+	// yarn init
```
knqyf263, line 2323:
```suggestion
	// apk add git
	// yarn init
```

knqyf263 (commented):
Is it possible to distinguish between v1 and v2? If `yarn.lock` is valid YAML in v2, we can use a yaml library to unmarshal it like `yaml.Unmarshal`.

antoinedeschenes:
@knqyf263 Sure, the file could be scanned for the __metadata line, or the presence of file `.yarnrc.yml` could be used (in fanal) to determine which parser version to use

merceyz:
The way to detect Yarn 2+ is to check the `yarn.lock` for `__metadata`, checking for `.yarnrc.yml` is not safe as it might not exist

PR #14, opened 2020-10-14
NuGet (packages.lock.json) parser.

Johannestegner:
This pull request includes a parser which parses lockfiles created by nuget/dotnet when using a lockfile.  
NuGet does not force (or auto) create a lock file, so a *.csproj file parser might also be a good idea, as a fallback.

As a first stage, this should be good enough though.


Johannestegner:
Reference: Issue with feature request can be found here https://github.com/aquasecurity/trivy/issues/681

simar7:
thanks for the PR @Johannestegner – I'll be taking a look at it and making comments as I go through it.

pkg/nuget/parse_testcase.go
```diff
@@ -0,0 +1,59 @@
+package nuget
```
simar7, line 1: I suggest moving these vars over to `pkg/nuget/parse_test.go` so it's easier to read while looking at the test.

simar7: Hmm - I see why you did this as the rest of project follows it. I take back what I said in that case :) My preference would still be to keep it together but I guess it doesn't hurt otherwise.

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+	unique := map[string]struct{}{}
+
+	// The first loop is the targets (.NETCoreApp,Version=v3.1)
+	// that means that it's not really a package, but there are dependencies
+	// under said target.
+	for _, targetContent := range lockFile.Dependencies {
+		// Here comes the 'real' dependencies.
+		for topPkgName, topPkgContent := range targetContent {
+			// And inside the dependency, there are another level of dependencies.
+			// But that's actually it, I swear!
```
simar7, line 50: It's pretty self documenting as code so you can take off the comments. Appreciate your commentary though!

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/json"
+	"fmt"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"io"
+)
+
+// The lockfile that nuget uses (packages.lock.json) works basically like this:
+// Each 'build type' (for example .NETCoreApp,Version=v3.1) have a set of dependencies.
+// Those dependencies can be 'Direct', 'Transitive' and 'Project'
+// (meaning of those are quite self explanatory I think). The dependency consists of a string
```
simar7, line 13:
```suggestion
// The dependency consists of a string
```

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, targetContent := range lockFile.Dependencies {
+		// Here comes the 'real' dependencies.
+		for topPkgName, topPkgContent := range targetContent {
+			// And inside the dependency, there are another level of dependencies.
+			// But that's actually it, I swear!
+			for pkgName, version := range topPkgContent.Dependencies {
+				symbol := fmt.Sprintf("%s@%s", pkgName, version)
+				if _, ok := unique[symbol]; ok {
+					continue
+				}
```
simar7, line 55: What happens if we have two transitive dependencies but of different versions, which one gets picked up? The first in the JSON? Is that desired behaviour?

simar7: I would extend that argument to any dependency type (transitive or otherwise). What happens in such a case?

Johannestegner: The symbol depends on both the package name and versions, so in a case where two versions exist as dependencies, it _should_ pick both up as different dependencies.  
I think that is the expected/intended way, right?

I will update with a testcase for this, and if I missunderstod the expected behavior, please let me know! :)

pkg/nuget/parse_testcase.go
Johannestegner: I felt the same when I wrote it, but yeah, wanted to keep the code as similar as possible to the other source :)

pkg/nuget/parse.go
Johannestegner: Update: Added test case for this. If it's not the intended behavior, please let me know and I'll fix it right away! :)

Johannestegner:
Thank you for the review @simar7 ! Fixed the issues you mentioned, let me know if there is anything else you wish me to fix! :)

pkg/nuget/parse.go
simar7: Thanks for adding the test case! This makes sense to me. If a project is using two different versions of dependencies somehow, I agree both should be taken into account. 

cc @knqyf263 would you have more insight into this?

simar7 (approved):
lgtm – @knqyf263 would you like to take a second look before we merge?

Johannestegner:
Not sure yet, but depending on how the version should be handled in Trivy (https://github.com/aquasecurity/trivy/pull/686#issuecomment-712509515) I guess that this part should use a similar approach to convert the version.
This could of course be done later on, but might want to do it in a one off :)

Johannestegner:
Actually, I guess this should _not_ be used in the dependency parser, but rather when checking in Trivy or similar.
I will add a test case for legacy packages though, so that it's in the codebase.

pkg/nuget/parse_testcase.go
knqyf263: That is because test cases are automatically generated by commands as follows because it is impossible to create the cases manually.
https://github.com/aquasecurity/go-dep-parser/blob/master/pkg/yarn/parse_testcase.go

It is too big and if we put it into a test file, it becomes difficult to distinguish between test cases and test logic. Also, the auto-generated file should be separated from the normal file. In such a case, it's common to define test cases in the dedicated file.

In this case, test cases are small enough. If we can maintain it manually, we can merge it into `parse_test.go`.

pkg/nuget/parse.go
knqyf263: It sounds fine.

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+// The following values should be parsed right away: resolved (resolved version),
+// type (this information might be worth having) and finally, it's dependencies.
+// The dependencies in this stage of the file are only a string map where the key is package name
+// and value is package version.
+
+type LockFile struct {
+	Version      int
+	Dependencies map[string]map[string]Dependency
+}
+
```
knqyf263, line 24: We probably should call it `Targets` according to the source code.
https://github.com/NuGet/NuGet.Client/blob/32cbb0ad13cf4b1abdb961b7ac111463e5f7b31a/src/NuGet.Core/NuGet.ProjectModel/ProjectLockFile/PackagesLockFileFormat.cs#L95

```suggestion
type LockFile struct {
	Version      int
	Targets map[string]Dependencies `json:"dependencies"`
}

type Dependencies map[string]Dependency
```

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+	ContentHash  string            `json:"contentHash"`
+}
+
+func Parse(r io.Reader) ([]types.Library, error) {
+	var lockFile LockFile
+	decoder := json.NewDecoder(r)
+	err := decoder.Decode(&lockFile)
+	if err != nil {
+		return nil, err
+	}
```
knqyf263, line 38: nit

```suggestion
	if err := decoder.Decode(&lockFile); err != nil {
		return nil, err
	}
```

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+	Dependencies map[string]string `json:"dependencies"`
+	ContentHash  string            `json:"contentHash"`
+}
+
+func Parse(r io.Reader) ([]types.Library, error) {
+	var lockFile LockFile
+	decoder := json.NewDecoder(r)
+	err := decoder.Decode(&lockFile)
+	if err != nil {
+		return nil, err
```
knqyf263, line 37: Could you wrap the error like the following?

```suggestion
		return nil, xerrors.Errorf("failed to decode package.lock.json: %w", err)
```

pkg/nuget/parse.go
```diff
@@ -0,0 +1,84 @@
...
+			// Else, if not unique, add the package to library list.
+			symbol := fmt.Sprintf("%s@%s", topPkgName, topPkgContent)
+			if _, ok := unique[symbol]; ok {
+				continue
+			}
+
+			libraries = append(libraries, types.Library{
+				Name:    topPkgName,
+				Version: topPkgContent.Resolved,
+			})
```
knqyf263, line 70: I hope the following change works.

```suggestion
			lib := types.Library{
				Name:    topPkgName,
				Version: topPkgContent.Resolved,
			})
			if _, ok := unique[lib]; ok {
				continue
			}

			libraries = append(libraries, lib)
```

pkg/nuget/parse_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+"os"
+"path"
+"sort"
+"strings"
+"testing"
+
+"github.com/aquasecurity/go-dep-parser/pkg/types"
+"github.com/kylelemons/godebug/pretty"
```
knqyf263: gofmt?

knqyf263 (commented):
Thanks for your contribution!

pkg/nuget/parse.go
Johannestegner: I missed those commands!
I generated a lockfile in one of my projects and removed some of the dependencies just to get a shorter list, but I'll create a new one via a similar command as the above!

knqyf263: It makes sense. The lock file looks broken since you removed some dependencies to make it shorter. It would be great if you write commands down to reproduce the lock files for testing.

Johannestegner: While fixing this, I re-read the whole nuget package resolving docs to check why I got it wrong. This made me rewrite the algorithm a bit...
NuGet will use the version of the package "closest" to the application, that is, it will resolve to use a package that is a direct dependency above using a sub-dependency. So if a package is installed by the user, that is the one that it will use.

What I changed was the following:

Instead of looping through the sub packages at the same time as the top level packages, I changed it to use two different loops.  
It will first go through the top packages, and then the sub-packages.
I also swapped the symbol for uniqueness to _only_ use the name of the package, and not both name and version, so that it will not add a package with two different versions.

I updated the test cases with new packages which I installed via a similar method of the one that was used in the other package managers (and included comments for reproducibility).

New code will be pushed when I'm done with the other changes you have asked for :)

pkg/nuget/parse.go
Johannestegner: This is no longer needed, as the uniqueness is now depending on the package name only, not version.

pkg/nuget/parse.go
knqyf263: I'm feeling like all we have to do is parse the top packages. The sub packages should exist on top level as well. If there are some version constraints in the sub packages, NuGet resolves it and pick up one version. But I'm not familiar with NuGet. If you think we need to parse the sub packages, could you tell me the document for it?

Johannestegner: Looking through the "complex" json file, I actually think you are correct, it seems like all sub-packages are in the file as `Transitive` dependencies in contrary to the ones installed by the user, which are `Direct`... I will double check to make sure that this is correct and if so, just drop the second pass of package traversal!

Johannestegner: From what I can see in the files I've tested with, you are correct!
So shortened the code a bit and updated the tests :)

Johannestegner:
Let me know if you want me to squash some (or all) commits :)

pkg/nuget/parse.go
knqyf263: >it seems like all sub-packages are in the file as Transitive dependencies in contrary to the ones installed by the user, which are Direct...

Yes, yes, that is exactly I wanted to say. If we pick versions in the sub packages, it would be false positive because v9.0.1 is actually not installed in the example I shared.

Thanks for your confirmation!

pkg/nuget/parse.go
knqyf263: What about different targets? I believe there is only one version in a target, but I'm wondering if the lock file contains multiple targets like ”.NETCoreApp,Version=v3" and ”.NETCoreApp,Version=v5", and they may have the same packages and different versions.

knqyf263:
I'll squash all commits when merging this PR, so you don't have to care about it. We should keep the history in the PR.

knqyf263:
I left one more comment, but it looks almost good! Thank you, again!

Johannestegner:
> I left one more comment, but it looks almost good! Thank you, again!

Good catch!

Updated the parser to use name + version again (tried with the code in an earlier note, but it didn't work out at all as I expected! hehe).
I removed the `complex` file/test and replaced it with a `multi` which contains targets for `net4.0;netstandard2.0;netstandard1.0;net35;net2.0`. Took me quite a while to find a package which required different dependencies!  
Now I think it is working just as it should! :)

knqyf263:
Thank you for adding multiple targets file! I've refactored a bit. It looks good to me now.

knqyf263 (approved):

PR #21, opened 2021-02-17
add go.sum support

### LINKED ISSUE #20: feat(golang): support go.sum
Parse `go.sum` and list all dependencies and versions. The signature of the function should be like `func Parse(r io.Reader) ([]types.Library, error)`.
Ref. https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/bundler/parse.go

ronaudinho:
resolves #20


pkg/gomod/parse_testcase.go
```diff
@@ -0,0 +1,388 @@
...
+	// docker run --name gomod --rm -it golang:1.15 bash
+	// export USER=gomod
+	// mkdir repo
+	// cd repo
+	// go mod init github.com/org/repo
+	// go get golang.org/x/xerrors
+	// go get github.com/urfave/cli
+	// go get github.com/stretchr/testify
+	// go get github.com/BurntSushi/toml
+	// go list -m all | awk 'NR>1 {sub(/^v/, "", $2); sub("-.*", "", $2); printf("{\""$1"\", \""$2"\"},\n")}'
```
ronaudinho, line 26: 1. using `go list -m` since it is a module-aware command and output is easier to parse than `cat go.sum`
2. ignoring first line since it is the name of the main module, which is ignored in `go.sum`

pkg/gomod/parse_testcase.go
```diff
@@ -0,0 +1,388 @@
+package gomod
+
+import "github.com/aquasecurity/go-dep-parser/pkg/types"
+
+var (
```
ronaudinho, line 5: not sure about repo to pick for test cases, let me know if other repos are preferred.

pkg/gomod/parse_testcase.go
knqyf263: Sounds great!

pkg/gomod/parse_test.go
```diff
@@ -0,0 +1,75 @@
...
+			libraries: GoModTrivy,
+		},
+	}
+
+	for _, v := range vectors {
+		t.Run(path.Base(v.file), func(t *testing.T) {
+			f, err := os.Open(v.file)
+			if err != nil {
+				t.Fatalf("Open() error: %v", err)
+			}
```
knqyf263, line 38:
```suggestion
			require.NoError(t, err)
```

Recently, we migrated to `github.com/stretchr/testify/assert` and `github.com/stretchr/testify/require`. Could you rewrite this test using them like the following?
https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/jar/parse_test.go

pkg/gomod/parse.go
```diff
@@ -0,0 +1,53 @@
...
+			Version: v,
+		})
+	}
+
+	return libs, nil
+}
+
+// parseSemVer parses semver from go.sum
+// ignoring pseudo-version
+func parseSemVer(v string) string {
```
knqyf263, line 41: I think we should keep pseudo-version. Let's say the latest git tag is `v1.0.0`. If there are some commits after `v1.0.0`, the version would be like `v1.0.1-0.20201117152358-0edc412565dc`. Then, the library will release `v1.0.1`. In that case, `v1.0.1` and `v1.0.1-0.20201117152358-0edc412565dc` are different and should be handled differently.

ronaudinho: that makes sense, updated to keep pseudo version

pkg/gomod/parse_test.go
ronaudinho: updated, rewrote the rest of the tests as well

pkg/gomod/parse.go
```diff
@@ -0,0 +1,47 @@
...
+	v = strings.TrimPrefix(v, "v")
+	vv := strings.Split(v, ".")
+	vv[2] = strings.Join(vv[2:], ".")
+	vv[2] = strings.TrimSuffix(vv[2], "/go.mod")
+
+	return strings.Join(vv[:3], ".")
```
knqyf263, line 46: Why are these `Split` and `Join` necessary? If they are needed, could you write examples as a comment?

pkg/gomod/parse_test.go
knqyf263: Great job! Thanks!

pkg/gomod/parse.go
ronaudinho: actually they are not necessary, updated

pkg/gomod/parse.go
```diff
@@ -0,0 +1,36 @@
...
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+		s := strings.Fields(line)
+		// go.sum records and sorts all non-major versions
+		// with the latest version as last entry
+		tmp[s[0]] = strings.TrimSuffix(strings.TrimPrefix(s[1], "v"), "/go.mod")
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
```
knqyf263, line 25:
```suggestion
		return nil, xerrors.Errorf("scan error: %w", err)
```
Could you wrap the error?

pkg/gomod/parse.go
```diff
@@ -0,0 +1,36 @@
...
+	"io"
+	"strings"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+)
+
+// Parse parses a go.sum file
+func Parse(r io.Reader) ([]types.Library, error) {
+	var libs []types.Library
+	tmp := make(map[string]string)
```
knqyf263, line 14:
```suggestion
	uniqLibs := make(map[string]string)
```
`tmp` is ambiguous.

knqyf263 (changes requested):

pkg/gomod/parse.go
ronaudinho: done, wrap errors in other pkgs as well

pkg/gomod/parse.go
ronaudinho: updated

pkg/gomod/parse.go
knqyf263: Thanks!

knqyf263 (approved):

knqyf263:
Thanks!

PR #22, opened 2021-02-20
support PKG-INFO and METADATA

ronaudinho:
Resolves https://github.com/aquasecurity/trivy/issues/716

`pkg/wheel/testdata` and `pkg/egg/testdata` contains many files, not sure if they are necessary for testing purpose since `wheel.Parse` and `egg.Parse` only parses file. More details on my thoughts in line comments.

### DIFF
--- a/pkg/egg/parse_test.go
+++ b/pkg/egg/parse_test.go
@@ -1,0 +1,84 @@
+package egg
+
+import (
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/xerrors"
+)
+
+func TestParse(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want []types.Library
+	}{
+		{
+			file: "testdata/egg_normal",
+			want: EggNormal,
+		},
+		{
+			file: "testdata/egg_many",
+			want: EggMany,
+		},
+		{
+			file: "testdata/egg_django",
+			want: EggDjango,
+		},
+	}
+
+	for _, v := range vectors {
+		t.Run(path.Base(v.file), func(t *testing.T) {
+			f, err := os.Open(v.file)
+			require.NoError(t, err)
+			got, err := walk(f)
+			require.NoError(t, err)
+
+			assert.Equal(t, v.want, got)
+		})
+	}
+}
+
+// walk is a test helper to traverse directories and look for METADATA files
+// given that Parse only parses METADATA, as traversing directories is done in fanal
+func walk(r io.Reader) ([]types.Library, error) {
+	var libs []types.Library
+	f, _ := r.(*os.File)
+	fi, err := os.Stat(f.Name())
+	if err != nil {
+		return nil, xerrors.Errorf("stat error: %w", err)
+	}
+	if !fi.IsDir() {
+		return nil, xerrors.New("not os.Dir")
+	}
+
+	walker := func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !info.IsDir() {
+			fileName := filepath.Base(path)
+			if fileName == "PKG-INFO" || strings.HasSuffix(fileName, "egg-info") {
+				f, err := os.Open(path)
+				if err != nil {
+					return err
+				}
+				lib, err := Parse(f)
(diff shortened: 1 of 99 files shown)


pkg/egg/parse_testcase.go
```diff
@@ -0,0 +1,70 @@
+package egg
+
+import "github.com/aquasecurity/go-dep-parser/pkg/types"
+
+var (
```
ronaudinho, line 5: since `egg-info` is old format, I haven't been able to find a way to diversify the test cases, so most of these are actually redundant.
it also seems that `trivy` picks up library based on the current language version? I ran a scan and `trivy` only checks for `python3` packages

pkg/wheel/parse.go
```diff
@@ -0,0 +1,29 @@
...
+	"io"
+	"net/textproto"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"golang.org/x/xerrors"
+)
+
+// Parse parses METADATA file for library Name and Version
+// currently does not parse its possible dependencies
+// https://packaging.python.org/specifications/core-metadata/#requires-dist-multiple-use
```
ronaudinho, line 14: should we scan for sub-dependencies? if so, I haven't found a method to pinpoint what current version is being used based on `Requires-Dist` in `METADATA` files, which I believe is necessary [here](https://github.com/aquasecurity/trivy/blob/a0cd5d70ae206ff9619c2d6d63a14830289fe520/pkg/detector/library/python/compare.go#L20)

pkg/egg/parse_test.go
```diff
@@ -0,0 +1,84 @@
...
+			require.NoError(t, err)
+
+			assert.Equal(t, v.want, got)
+		})
+	}
+}
+
+// walk is a test helper to traverse directories and look for METADATA files
+// given that Parse only parses METADATA, as traversing directories is done in fanal
+func walk(r io.Reader) ([]types.Library, error) {
```
ronaudinho, line 50: this and [this](https://github.com/aquasecurity/go-dep-parser/pull/22/files?file-filters%5B%5D=.go#diff-fa141eeaea210ac278122d59e1f38d922d3ee9436afdfcc73774b64b8757f6cdR49) might not be necessary here?

ronaudinho:
as an addition, current way of parsing `wheel` and `egg` will also affect fanal [here](https://github.com/aquasecurity/fanal/blob/98d76e2b3b9605fd0e3cf9b1fc380e7d94ad2af2/cmd/fanal/main.go#L181), as it treats each file as separate application.

pkg/egg/parse_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"golang.org/x/xerrors"
+)
+
+func TestParse(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want []types.Library
+	}{
+		{
+			file: "testdata/egg_normal",
```
knqyf263, line 23: We should pass a single file instead of a directory.

pkg/egg/parse_test.go
```diff
@@ -0,0 +1,84 @@
...
+			want: EggNormal,
```
knqyf263, line 24: The expected value should also be a single package name/version.

knqyf263 (changes requested):
@ronaudinho Thanks! First of all, can we split this PR into two, egg and wheel? I'd like to review them separately.

In the case of lock files, a single file contains some dependencies, so expected values in unit tests must be a list of dependencies. But in these cases, a single file contains only 1 dependency. So, we don't have to pass multiple files, but a single file.

In addition, if the file format is a bit different from other files (in other words, it increases coverage), we should add it to test cases. If it's not the case, we don't have to add it.

pkg/wheel/parse.go
knqyf263: If I understand correctly, we don't have to extract sub-dependencies. If they are actually necessary, those wheel would exist in the same system.

pkg/wheel/parse_test.go
```diff
@@ -0,0 +1,80 @@
...
+			file: "testdata/wheel_django",
+			want: WheelDjango,
+		},
+	}
+
+	for _, v := range vectors {
+		t.Run(path.Base(v.file), func(t *testing.T) {
+			f, err := os.Open(v.file)
+			require.NoError(t, err)
+			got, err := walk(f)
```
knqyf263, line 39: So, `walk` is unnecessary here. We can simply pass a single file to `Parse`. I believe it would remove most of METADATA files under `testdata`.

ronaudinho:
Closing in favor of #23 and #24

PR #27, opened 2021-04-20
fix: add an option to override http.Client used by jar parser

arajkumar:
This would enable the embedder of this library to decide whether to allow or disallow network calls to maven central. Those who don't want the remote call shall pass dummy client.

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

### DIFF
--- a/pkg/jar/parse.go
+++ b/pkg/jar/parse.go
@@ -36,6 +36,7 @@ var (
 
 type conf struct {
 	baseURL      string
+	client       *http.Client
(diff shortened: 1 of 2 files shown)


knqyf263 (approved):

knqyf263:
Cool! Thanks.

arajkumar:
Thanks for this awesome package :) Additional thanks for merging this 🙏 .

PR #28, opened 2021-04-24
fix(jar): return errors other than no artifact found error

masahiro331:


pkg/jar/parse.go
```diff
@@ -418,7 +421,7 @@ func searchBySHA1(c conf, data []byte) (properties, error) {
 	}
 
 	if len(res.Response.Docs) == 0 {
-		return properties{}, xerrors.Errorf("no artifact found: %s", digest)
+		return properties{}, xerrors.Errorf("digest %s: %w", digest, ArtifactNotFoundErr)
```
knqyf263, line 424: We should also return an error if searchBySHA1 returns an error other than ArtifactNotFoundErr.

knqyf263 (approved):

PR #34, opened 2021-05-26
fix(gobinary): check first column to determine meaning

ebati:
I updated the gobinary line checking logic to understand replace directive("=>") and also added ability to understand lines without hash values.

I do not know why i do not get hash values when i build binary in a container with vendor folder but you can find the steps in [this repo](https://github.com/ebati/trivy-mod-parse). The test binary (replace.elf) is also from the code in that repo.

Related to aquasecurity/trivy/issues/980

### DIFF
--- a/pkg/gobinary/parse.go
+++ b/pkg/gobinary/parse.go
@@ -6,6 +6,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"io"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


masahiro331:
LGTM

masahiro331 (approved):

masahiro331:
LGTM!

pkg/gobinary/parse.go
```diff
@@ -29,14 +29,29 @@ func Parse(r io.Reader) ([]types.Library, error) {
...
-		})
+		switch ss[0] {
+		case "dep":
+			libs = append(libs, types.Library{
+				Name:    ss[1],
+				Version: ss[2],
+			})
+		case "=>":
+			// replace replaces the previous entry
+			prev := len(libs) - 1
```
masahiro331, line 47: IMHO.

It is not possible by golang elf design, but len​​(libs) may return 0.

ebati: Should i return an error or skip that line only or assume this is not a valid go binary and `return nil, nil`?

knqyf263: I prefer returning an error here.

knqyf263 (approved):

knqyf263:
@ebati Thanks a lot!

PR #35, opened 2021-06-21
Parse java php nodejs binary version

laurentdelosieresmano:
Hello,

This PR parses the versions of standalone binaries (PHP / Java / Node.js) that are extracted from tarballs and not packages. Let me know if you wish to add more conditions to avoid false positives. Unfortunately, the filename was not passed to the "Parse()" function; otherwise, I could have added an extra condition on the filename (e.g. if the file starts with "java", or "node").

This resolves https://github.com/aquasecurity/trivy/issues/1064.

Best,
Laurent


bhbitter:
What is the status of merging this feature? I have the same issue where trivy doesn't report the vulnerable version of nodejs.

PR #36, opened 2021-06-27
changing base url to https

sherif84:
proposing to change the base url to https for increased security and TLS encryption


knqyf263 (approved):

knqyf263:
Thanks!

PR #37, opened 2021-07-07
Enhance nuget parser for packages.config

TauseefMalik:
This commit adds the ability to parse packages.config project file in Dotnet and identifies the packages and versions. Also, added
unit tests for the same.


knqyf263:
You should have another parser.

pkg/nuget/parse.go
```diff
@@ -1,54 +0,0 @@
-package nuget
```
knqyf263, line 1: Could you try `git mv`? I want git to know this is renaming.

pkg/nugetconfig/parse_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+)
+
+func TestParse(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want []types.Library
+	}{
```
knqyf263, line 20: Could you rewrite the test as below?
https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/gobinary/parse_test.go#L14-L19

pkg/nugetconfig/parse_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{
+			file: "testdata/packages.config",
+			want: NuGetPackagesConfig,
+		},
```
knqyf263, line 24: You should sad path.

pkg/nugetconfig/parse_testcase.go
```diff
@@ -0,0 +1,13 @@
...
+import "github.com/aquasecurity/go-dep-parser/pkg/types"
+
+var (
+	NuGetPackagesConfig = []types.Library{
+		{"Microsoft.AspNet.WebApi", "5.2.2"},
+		{"Microsoft.AspNet.WebApi.Client", "5.2.2"},
+		{"Microsoft.AspNet.WebApi.Core", "5.2.2"},
+		{"Microsoft.AspNet.WebApi.WebHost", "5.2.2"},
+		{"Newtonsoft.Json", "6.0.4"},
+	}
```
knqyf263, line 12: Could you write down how to generate `packages.config`? You can refer to other parsers.

pkg/nugetconfig/parse.go
```diff
@@ -0,0 +1,54 @@
...
+type Dependency struct {
+	Type     string
+	Resolved string
+}
+
+type cfgPackageReference struct {
+	XMLName         xml.Name `xml:"package"`
+	TargetFramework string   `xml:"targetFramework,attr"`
+	Version         string   `xml:"version,attr"`
+	ID              string   `xml:"id,attr"`
```
knqyf263, line 21: We should add `developmentDependency` and skip it when it is true.
https://docs.microsoft.com/en-us/nuget/reference/packages-config

knqyf263 (changes requested):

pkg/nugetconfig/parse_test.go
```diff
@@ -0,0 +1,54 @@
+package nugetconfig
```
knqyf263, line 1: It is better to use another package if possible.

```suggestion
package nugetconfig_test
```

TauseefMalik: Parse method exists in nugetconfig package, hence kept it as is.

pkg/nugetconfig/parse_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestParse(t *testing.T) {
+	tests := []struct {
+		name      string // Test input file
+		inputFile string
+		want      []types.Library
+		wantErr   string
+	}{
+		{
+			name:      "Config",
```
TauseefMalik, line 21: Since, packages.config is legacy format, it is supported when creating .net project in windows with older .NET framework using Visual studio IDE. e.g. https://github.com/microsoft/AzureBot/blob/master/AzureBot/packages.config

pkg/nugetlock/parse.go
```diff
@@ -0,0 +1,54 @@
+package nugetlock
```
TauseefMalik, line 1: tried to rename with git mv, did not work.

pkg/nugetlock/parse_testcase.go
```diff
@@ -0,0 +1,98 @@
+package nugetlock
```
TauseefMalik, line 1: tried to rename with git mv, did not work.

pkg/nugetconfig/parse.go
```diff
@@ -0,0 +1,57 @@
...
+type Dependency struct {
+	Type     string
+	Resolved string
+}
+
+type cfgPackageReference struct {
+	XMLName         xml.Name `xml:"package"`
+	TargetFramework string   `xml:"targetFramework,attr"`
+	Version         string   `xml:"version,attr"`
+	DevDependency   string   `xml:"developmentDependency,attr"`
```
TauseefMalik, line 22: Added devDependency attribute check

pkg/nugetconfig/parse_test.go
knqyf263: What do you mean?

pkg/nugetlock/parse_testcase.go
knqyf263: It's done.
https://github.com/aquasecurity/go-dep-parser/pull/42

knqyf263: Rebase or merge the above PR

pkg/nugetconfig/parse_test.go
knqyf263: The document says `packages.config` is created when the first NuGet operation is run. Is it really impossible to generate it?

>If used, packages.config is typically located in a project root. It's automatically created when the first NuGet operation is run, but can also be created manually before running any commands such as nuget restore.

https://docs.microsoft.com/en-us/nuget/reference/packages-config

pkg/nugetconfig/parse.go
knqyf263: Any reason you cannot use boolean?

knqyf263 (changes requested):
Also, the test is failing. I recommend running tests in your local at first.

pkg/nugetconfig/parse.go
TauseefMalik: done

pkg/nugetconfig/parse_test.go
TauseefMalik: done

pkg/nugetconfig/parse_test.go
TauseefMalik: Here's what I mean:
https://stackoverflow.com/questions/42293369/missing-packages-config-file
It is little tricky to create using new dotnet cli commands but can be added easily through Visual studio IDE with some tweaks to it's package manager settings.

TauseefMalik: image

pkg/nugetconfig/parse.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"golang.org/x/xerrors"
+)
+
+type Dependencies map[string]Dependency
+
+type Dependency struct {
+	Type     string
+	Resolved string
+}
```
knqyf263, line 16: Looks like they are not used.

knqyf263 (approved):

PR #17, opened 2021-01-30
feat: support gemspec

danfaizer:
`go-dep-parse` extracts dependencies of gemspec files.
This PR is related with [issue](https://github.com/aquasecurity/trivy/issues/700).
Feel free to modify, comment, close or suggest modifications.

PR #47, opened 2021-08-03
Added new gemspec parser.

ankk13:
Parser reads the file ending with .gemspec. 
Refers value in it or the filepath to get name and version of the package.


pkg/gemspec/parse_testcase.go
```diff
@@ -0,0 +1,18 @@
...
+	}
+	JsonGem = []types.Library{
+		{"json", "2.3.0", "Ruby"},
+	}
+	JsonVersionGem = []types.Library{
+		{"json", "VERSION", "Ruby"},
+	}
+	JsonNameGem = []types.Library{
+		{"$json", "2.3.0", "Ruby"},
+	}
```
knqyf263, line 17: You can merge them into parse_test.go.

pkg/gemspec/testdata/json-2.3.0.gemspec
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env jruby
+require "rubygems"
+
+spec = Gem::Specification.new do |s|
+  s.name = "$json"
```
knqyf263, line 5: Could you tell me the spec of the name field? Any documentation? I want to know why `$` is accepted here.

pkg/gemspec/parse.go
```diff
@@ -0,0 +1,104 @@
...
+	"unicode/utf8"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"golang.org/x/xerrors"
+)
+
+var (
+	bwQuotes         = regexp.MustCompile(`['|"](.*?)['|"]`)
+	gemSpec          = ".gemspec"
+	gemStr           = ".name = "
```
knqyf263, line 18: Spaces are mandatory? I wonder if `.name="foo"` is allowed.

pkg/gemspec/parse.go
```diff
@@ -0,0 +1,104 @@
...
+	gemStrVer        = ".version = "
+	gemStrLic        = ".licenses = "
+	packageNameRegEx = regexp.MustCompile(`^[a-zA-Z0-9_\-]+$`)
+	versionRegEx     = regexp.MustCompile(`^[0-9.]+$`)
+	fileNameVersion  = regexp.MustCompile(`^([\w_.\-]+)-([0-9.]+)$`)
+)
+
+func Parse(r io.Reader, filePath string) ([]types.Library, error) {
```
knqyf263, line 26: If it always returns a single library, you don't have to use []types.Library.

```suggestion
func Parse(r io.Reader, filePath string) (types.Library, error) {
```

pkg/gemspec/testdata/json-2.3.0.gemspec
ankk13: you can refer https://guides.rubygems.org/specification-reference/#name. It suggests formal syntax.
To explore, I tried below commented regexes on image kankush/ruby:v2. This image has famous ruby packages installed.
I did not miss the name from any gemspec file. But I still think regexes should be used, as this is not strict format. 

$json is no different than simple package name. So should be considered.

pkg/gemspec/parse.go
ankk13: addressed in above comment

knqyf263:
Close this PR in favor of https://github.com/aquasecurity/go-dep-parser/pull/17

PR #48, opened 2021-08-03
feat: add wordpress parser

ankk13:
wordpress is a well used php resource. But it do not use composer dependency framework.
To support wordpress detection, composer parser is modified to support /wp-includes/version.php file.


knqyf263 (commented):
I don't think this should be included in the composer parser

pkg/composer/parse.go
```diff
@@ -33,3 +36,25 @@ func Parse(r io.Reader) ([]types.Library, error) {
...
+	// If wordpress file, open file and
+	// find line with content
+	// $wp_version = '<WORDPRESS_VERSION>';
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if !strings.HasPrefix(line, "$wp_version") {
+			continue
+		}
+		wpVersionRegex, _ := regexp.Compile("'(.*?)'")
```
knqyf263, line 51: `Compile` should be done outside.

pkg/composer/parse_test.go
```diff
@@ -46,3 +46,27 @@ func TestParse(t *testing.T) {
...
+
+func TestParseWordPress(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want []types.Library
+	}{
+		{
+			file: "testdata/wp-includes/version.php",
+			want: Wordpress,
+		},
```
knqyf263, line 58: add more cases

pkg/composer/parse.go
```diff
@@ -33,3 +36,25 @@ func Parse(r io.Reader) ([]types.Library, error) {
...
+		version := strings.Trim(wpVersionRegex.FindString(line), "'")
+		libs = append(libs, types.Library{
+			Name:    "wordpress",
+			Version: version,
+			License: "",
+		})
```
knqyf263, line 57: Is it possible that multiple versions are defined? If no, we should break the loop.

pkg/composer/parse.go
```diff
@@ -33,3 +36,25 @@ func Parse(r io.Reader) ([]types.Library, error) {
 	}
 	return libs, nil
 }
+
+func ParseWordPress(r io.Reader) ([]types.Library, error) {
```
knqyf263, line 40: If multiple versions are not allowed, we should return a single object.
```suggestion
func ParseWordPress(r io.Reader) (types.Library, error) {
```

pkg/composer/parse_test.go
masahiro331: e.g. 
```
// $wp_version = '<WORDPRESS_VERSION>'; // skip 
/**
 * $wp_version = '<WORDPRESS_VERSION>';  // skip 
**/
$wp_version = '<WORDPRESS_VERSION>';
```

pkg/php/wordpress/parse.go
```diff
@@ -0,0 +1,65 @@
...
+	// find line with content
+	// $wp_version = '<WORDPRESS_VERSION>';
+	content, err := readAll(r)
+	if err != nil {
+		return lib, xerrors.Errorf("decode error: %w", err)
+	}
+
+	contentStr := string(commentRegex.ReplaceAll(content, []byte{}))
+
+	matches := wpVersionRegex.FindAllStringSubmatch(contentStr, 3)
```
knqyf263, line 36: I think we can process line by line and simplify regular expressions.

```suggestion
	var version string
	scanner := bufio.NewScanner(r)
	for scanner.Scan() {
		line := scanner.Text()
		
		// It might include $wp_version_something
		if strings.HasPrefix(line, "$wp_version") {
			continue
		}
	
		ss := strings.Split(line, "=")
		if len(ss) != 2 || strings.TrimSpace(ss[0]) != "$wp_version" {
			continue
		}
		version = strings.Trim(ss[1], `'";`)
		break
	}
	if err := scanner.Err(); err != nil {
		return types.Library{}, xerrors.New("version.php could not be parsed")
	}
```

pkg/php/wordpress/parse_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestParseWordPress(t *testing.T) {
+	tests := []struct {
+		file    string // Test input file
+		want    types.Library
+		wantErr string
+	}{
+		{
+			file:    "testdata/version.php",
+			want:    types.Library{"wordpress", "4.9.4-alpha", ""},
```
knqyf263, line 22:
```suggestion
			want:    types.Library{
				Name: "wordpress", 
				Version: "4.9.4-alpha",
			},
```

pkg/php/wordpress/parse_test.go
```diff
@@ -0,0 +1,47 @@
...
+			wantErr: "",
+		},
+		{
+			file:    "testdata/versionFail.php",
+			want:    *new(types.Library),
```
knqyf263, line 27: I think you can simply omit this field.
```suggestion
```

knqyf263 (changes requested):
Also, we should move the directory to "framework" or something like that.

pkg/frameworks/wordpress/parse.go
```diff
@@ -0,0 +1,64 @@
...
+
+	var version string
+	isComment := false
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.HasPrefix(line, "/*") {
+			continue
+		}
```
knqyf263, line 32: I think you have to set isCommnet=true.

pkg/frameworks/wordpress/testdata/version.php
```diff
@@ -0,0 +1,50 @@
...
+ * Contains version information for the current WordPress release.
+ *
+ * @package WordPress
+ * @since 1.1.0
+ */
+
+ // $wp_version = '4.5.6-alpha';
+
+ /* 
+    $wp_version = '4.5.6-alpha';
```
knqyf263, line 14: This case will fail.

```suggestion
$wp_version = '4.5.6-alpha';
```

knqyf263 (changes requested):

pkg/frameworks/wordpress/parse.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"golang.org/x/xerrors"
+)
+
+var (
+	wpVersionRegex = regexp.MustCompile(`(\$wp_version\s*=\s*['"])(.*)(['"]\s*;)`)
+	commentRegex   = regexp.MustCompile(`(/\*([^*]|[\r\n]|(\*+([^*/]|[\r\n])))*\*+/)|(//.*)`)
+)
```
knqyf263, line 16: unnecessary

knqyf263 (approved):

PR #49, opened 2021-08-03
feat(nodejs): add package.json parser for npm resource

ankk13:
New parser reads package.json files and collect package name, package version and license.
We do not intend to read dependent package's from this json file.


pkg/packagejson/parse.go
```diff
@@ -0,0 +1,37 @@
...
+func Parse(r io.Reader) ([]types.Library, error) {
+	libs := []types.Library{}
+	var data packageDotJSON
+	err := json.NewDecoder(r).Decode(&data)
+	if err != nil {
+		return libs, xerrors.Errorf("decode error: %w", err)
+	}
+
+	// the license isn't always a string, so only take it if it is a string
+	license, _ := data.License.(string)
```
masahiro331, line 27: I have a few questions.

License specification.
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license

There are five main ways to express a license.

1. One license
```
{
  "license" : "BSD-3-Clause"
}
```

2. SPDX license expression syntax (version 2.0)
```
{
  "license" : "(ISC OR GPL-3.0)"
}
```

3. External file reference 
```
{
  "license" : "SEE LICENSE IN <filename>"
}
```

4. Legacy format 
```
// Not valid metadata
{
  "license" : {
    "type" : "ISC",
    "url" : "https://opensource.org/licenses/ISC"
  }
}
// Not valid metadata
{
  "licenses" : [
    {
      "type": "MIT",
      "url": "https://www.opensource.org/licenses/mit-license.php"
    },
    {
      "type": "Apache-2.0",
      "url": "https://opensource.org/licenses/apache2.0.php"
    }
  ]
}
```

Is this pull request intended to address specifications 1 and 2?
As for 3, I think we could either skip it or parse.
As for 4, I'd like to discuss it, but is it possible to convert it to SPDX?

pkg/packagejson/parse_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+)
+
+func TestParse(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want []types.Library
+	}{
+		{
```
masahiro331, line 21: Could you add a test with a blank version or name?

pkg/packagejson/parse.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type packageDotJSON struct {
+	Name    string      `json:"name"`
+	Version string      `json:"version"`
+	SHA     string      `json:"_shasum"`
+	License interface{} `json:"license"`
+}
+
+func Parse(r io.Reader) ([]types.Library, error) {
```
masahiro331, line 18: Is it possible that there are multiple libraries?
There is no need to implement a specific Interface, so a single return value is fine.

pkg/packagejson/parse.go
ankk13: For case 1, 2 & 3 currently complete string value will be picked. 
Legacy format is not supported, but will update code to support it.

pkg/nodejs/packagejson/parse.go
```diff
@@ -0,0 +1,58 @@
...
+	return lib, nil
+}
+
+func parseLicense(val interface{}) string {
+	license, ok := val.(string)
+	if ok {
+		license = strings.ReplaceAll(license, " OR ", " | ")
+		license = strings.ReplaceAll(license, " AND ", ", ")
+		license = strings.ReplaceAll(license, " WITH ", " with ")
+		license = strings.TrimLeft(license, "SEE LICENSE IN ")
```
knqyf263, line 50: Could you tell me why we need this kind of manipulation?

pkg/nodejs/packagejson/parse.go
```diff
@@ -0,0 +1,58 @@
...
+			Name:    data.Name,
+			Version: data.Version,
+			License: license,
+		}
+	}
+	return lib, nil
+}
+
+func parseLicense(val interface{}) string {
+	license, ok := val.(string)
```
knqyf263, line 45:
```suggestion
	switch v := val.(type) {
	case string:
		// TODO
	case legacyLicense:
		// TODO
```

pkg/nodejs/packagejson/parse.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"golang.org/x/xerrors"
+)
+
+type packageJSON struct {
+	Name    string      `json:"name"`
+	Version string      `json:"version"`
+	SHA     string      `json:"_shasum"`
```
knqyf263, line 15: Is this necessary?

pkg/nodejs/packagejson/parse_test.go
```diff
@@ -0,0 +1,119 @@
+package packagejson
```
knqyf263, line 1:
```suggestion
package packagejson_test
```

knqyf263 (changes requested):

pkg/nodejs/packagejson/parse.go
ankk13: so need as such. Just for other parsers, we are using separators like comma. We can put the string as is.

pkg/nodejs/packagejson/parse.go
ankk13: Nope. I will remove it.

pkg/nodejs/packagejson/parse_test.go
ankk13: For all other unit tests in the project, we did not use similar package naming. I was expecting naming to be in sync.

pkg/packagejson/parse.go
masahiro331: In the case of "SEE IN $filename", is it OK to just use the string as is, instead of going to reference it?

pkg/nodejs/packagejson/parse.go
ankk13: Done

knqyf263 (changes requested):
You moved the test file under testdata by mistake.

pkg/nodejs/packagejson/testdata/parse_test.go
```diff
@@ -0,0 +1,73 @@
...
+			// docker run --name composer --rm -it node:12-alpine sh
+			// npm init --force
+			// npm install --save promise jquery
+			// npm ls | grep -E -o "\S+@\S+" | awk -F@ 'NR>0 {printf("{\""$1"\", \""$2"\"},\n")}'
+			want: types.Library{
+				Name:    "bootstrap",
+				Version: "5.0.2",
+				License: "MIT",
+			},
+		},
```
knqyf263, line 32: You should add a test case for the legacy license.

pkg/nodejs/packagejson/testdata/parse_test.go
```diff
@@ -0,0 +1,73 @@
...
+			expected: "MIT",
+		},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			got := packagejson.ParseLicense(c.value)
+			assert.Equal(t, c.expected, got)
+		})
+	}
+}
```
knqyf263, line 73: This test is unnecessary.

pkg/nodejs/packagejson/testdata/parse_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/aquasecurity/go-dep-parser/pkg/nodejs/packagejson"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+)
+
+func TestParse(t *testing.T) {
+	vectors := []struct {
+		file string // Test input file
+		want types.Library
+	}{
```
knqyf263, line 19: I fixed this test before, but you did force push and my change was overwritten. Can you refer to this test?
https://github.com/aquasecurity/go-dep-parser/blob/dda1673181319dd5f216ba4324e75ab98bdc296f/pkg/ruby/gemspec/parse_test.go#L15-L20

knqyf263 (changes requested):

pkg/nodejs/packagejson/testdata/parse_test.go
ankk13: Updated similar to mentioned testcases.

knqyf263 (approved):

PR #50, opened 2021-08-13
python: merge egg and wheel parsers

ankk13:


pkg/pymetadata/parse.go
```diff
@@ -9,9 +9,6 @@ import (
 	"golang.org/x/xerrors"
 )
 
-// Parse parses METADATA file for library Name and Version
-// currently does not parse its possible dependencies
-// https://packaging.python.org/specifications/core-metadata/#requires-dist-multiple-use
```
knqyf263, line 14: We should leave comments about what this parser does.

pkg/pymetadata/parse.go
```diff
@@ -9,9 +9,6 @@ import (
...
 func Parse(r io.Reader) ([]types.Library, error) {
```
knqyf263, line 12: We can return a single value.

```suggestion
func Parse(r io.Reader) (types.Library, error) {
```

pkg/pymetadata/parse_test.go
```diff
@@ -0,0 +1,103 @@
+package pymetadata
```
knqyf263, line 1:
```suggestion
package pymetadata_test
```

pkg/pymetadata/parse_test.go
```diff
@@ -0,0 +1,103 @@
...
+			},
+		},
+	}
+
+	for _, v := range vectors {
+		t.Run(path.Base(v.file), func(t *testing.T) {
+			f, err := os.Open(v.file)
+			require.NoError(t, err)
+
+			got, err := Parse(f)
```
knqyf263, line 97:
```suggestion
			got, err := pymetadata.Parse(f)
```

knqyf263 (approved):

Issue #65, opened 2021-11-14
Support pnpm-lock.yaml


knqyf263:
## Description
Parse pnpm-lock.yaml so as to extract package names/versions.
https://github.com/pnpm/spec/blob/master/lockfile/5.md

Example:
https://github.com/pnpm/pnpm/blob/main/pnpm-lock.yaml

## Issue
https://github.com/aquasecurity/trivy/issues/1200

WorldVirus:
Hi, 
I tried to push a branch for resolving this issue, but I have not access to push new branch

knqyf263:
You can search how to make Pull Request on GitHub.

WorldVirus:
And for creating new PR, I need to push my branch in repository, but i have 

```
remote: Permission to aquasecurity/go-dep-parser.git denied to WorldVirus.
fatal: unable to access 'https://github.com/aquasecurity/go-dep-parser/': The requested URL returned error: 403
```

weyert:
You need fork the repo and then push the branch to your repo and then after you pushed it should be able to make a PR. 
See also: https://makeapullrequest.com

WorldVirus:
I opened new pr, but could not update the reviewers

eranshapira:
@WorldVirus any intention on taking it further? I might take it if not.

itaysk:
@eranshapira are you actually going to take it?

weyert:
In a different project I have been working on a PNPM lock file parsing, you can find it here:
https://gitlab.com/weyert-tapico/gemnasium/-/commit/4fa4879e1fd5f3f8c26e7e3bfd64e5ecb9bf69dc

Feel free to use it

knqyf263:
Cool! Do you mind if we ask you to contribute?

PR #66, opened 2021-12-07
support pnpm parsing

WorldVirus:
https://github.com/aquasecurity/go-dep-parser/issues/65


knqyf263:
Thanks for the PR! Could you sign our CLA?

WorldVirus:
Yes, I've signed

WorldVirus:
Is it PR actual ?

masahiro331:
@WorldVirus 
https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/nodejs/npm/parse.go#L35-L37
Other parsers skip dev packages. 

Is it possible to use a defined struct instead of using an interface?

pkg/nodejs/pnpm/testdata/pnpm-lock-normal.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+dependencies:
+  jquery: 3.6.0
+  promise: 8.1.0
+
+packages:
+
+  /asap/2.0.6:
+    resolution: {integrity: sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=}
+    dev: false
```
eranshapira, line 15:
```suggestion
    dev: true
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -0,0 +1,102 @@
...
+
+	for packageInfo, packAdditionalInfo := range listPackages {
+		version, name := takePackageInfo(packageInfo)
+		curPckg := packAdditionalInfo.(map[string]interface{})
+		if val, ok := curPckg["version"]; ok {
+			version = val.(string)
+		}
+		if val, ok := curPckg["name"]; ok {
+			name = val.(string)
+		}
```
eranshapira, line 65:
```suggestion
		if val, ok := curPckg["name"]; ok {
			name = val.(string)
		}
		if val, ok := curPckg["dev"]; ok {
			dev = val.(string)
		        if dev == "true" {
		                continue
		        }
		}		
```

pkg/nodejs/pnpm/parse_testcase.go
```diff
@@ -0,0 +1,327 @@
...
+var (
+	// docker run --name node --rm -it node:12-alpine sh
+	// npm install -g pnpm
+	// pnpm add promise jquery
+	// pnpm list -depth 10 | grep -E -o "\S+\s+[0-9]+(\.[0-9]+)+$" | awk '{printf("{\""$1"\", \""$2"\",\"\"},\n")}' | sort -u
+	// have to hardcode value of depth flag because of there is a problem with output dependencies  https://github.com/pnpm/pnpm/issues/3583
+	pnpmNormal = []types.Library{
+		{"jquery", "3.6.0", ""},
+		{"promise", "8.1.0", ""},
+		{"asap", "2.0.6", ""},
```
eranshapira, line 16:
```suggestion
```

eranshapira (changes requested):
add support to ignore dev packages and add test for dev packages

knqyf263:
@WorldVirus Is this still relevant for you? We can take over the task.

eranshapira:
@knqyf263 this is actually relevant for me. i'm looking into making use of pnpm down the road. (for some of Aqua's products)

knqyf263:
@eranshapira OK, so can you take over it?

eranshapira:
@knqyf263 i just started working here a week ago, so it might take some time. but basically I wrote a review with the proposed changes (ignoring dev packages and testing it) 
Regarding using struct instead of interface, I can do it down the road.

knqyf263:
>Regarding using struct instead of interface, I can do it down the road.

Yes, struct is preferred.

knqyf263:
In favor of https://github.com/aquasecurity/go-dep-parser/pull/106

PR #72, opened 2021-12-30
Adding library location logging.

Freydal:
After the log4j fiasco we utilized Trivy to address vulnerabilities.  Some java applications were triggering CVEs despite efforts to override versions with dependencyManagement.  I took a little time to dig down in Trivy and this is the minimal change solution I used to find more information to lead to resolving CVE tied to fat jars being included in our builds.

Most of the culprits were ehcache or test jars not properly scoped.  A more sophisticated solution to provide that information for this case in the table output may be warranted, but I thought I'd share my simple solution.

aquasecurity/trivy#1141

### DIFF
--- a/pkg/python/pip/parse.go
+++ b/pkg/python/pip/parse.go
@@ -13,6 +13,7 @@ import (
 const (
 	commentMarker string = "#"
 	endColon      string = ";"
+	hashMarker    string = "--"
 )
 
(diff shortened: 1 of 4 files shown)


Freydal:
Sorry emails didn't match with that machine and this was closed when working to fix commit history.

PR #76, opened 2022-01-13
Add support for go.mod as well as go.sum

### LINKED ISSUE #75: Golang mod parser reports version that are not used in application
I'm currently looking into why the parser reports versions being used that are not even in my application.
I noticed it's because the issues were being reported in dependencies of dependencies (sometimes even further down the tree), the code was not reachable at all, so the code is not included in the application, yet it reports it as a vulnerability.

Now I do get this, it's impossible for the parser to see if a dependency is actually being used. The good thing is that Go automatically does this for you already. It does this by adding all dependencies into the go.mod file, even the transitive dependencies (marked with `//indirect`), as long as there is an import somewhere, it will be added to the go.mod file.

The Go blog says this about the go.sum file:
> The go command uses the go.sum file to ensure that future downloads of these modules retrieve the same bits as the first download, to ensure the modules your project depends on do not change unexpectedly, whether for malicious, accidental, or other reasons.

So I don't really see why go.sum is used here, and not go.mod. Can we change the parser to use go.mod? 
When you execute `go mod edit -json` you will get a JSON representation of the `go.mod` file that's very clean and easy to parse.

CC @rahul2393, since you have implemented this dep parser in #21

jerbob92:
Fixes #75


knqyf263:
Hi @jerbob92, I refactored a bit. Please let me know if I did something wrong.

knqyf263 (approved):

PR #82, opened 2022-02-24
fix(pip): handling a package name with a hyphen

### LINKED ISSUE #81: `go-dep-parser` handles incorrectly a package name with a hyphen
there is a problem if a package name contains a hyphen.

`go-dep-parser` doesn't detect next packages for python:
```
oauth2-client==4.0.0
python-gitlab==2.0.0
```
Reason:
`rStripByKey(line, hashMarker)`  cuts incorrectly such names.

https://github.com/aquasecurity/go-dep-parser/blob/4a30ebc4b509209c8e61561930cd760c28a8201d/pkg/python/pip/parse.go#L44-L49

afdesk:
Changed getting the index of the first instance of substr instead of the index of the first instance of any Unicode code point from chars.

Fixes #81

### DIFF
--- a/pkg/python/pip/parse.go
+++ b/pkg/python/pip/parse.go
@@ -42,7 +42,7 @@ func Parse(r io.Reader) ([]types.Library, error) {
 }
 
 func rStripByKey(line string, key string) string {
-	if pos := strings.IndexAny(line, key); pos >= 0 {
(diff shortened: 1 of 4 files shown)


knqyf263 (approved):

knqyf263:
Great!

PR #87, opened 2022-03-16
refactor(go/binary): use debug/buildinfo to find dependencies

ebati:
I am not sure if we have an API compatibility promise but i didn't remove error definitions to keep the compatibility. If you would like to keep the current version i can add it for version < go1.18


knqyf263:
Thanks for your contribution! Could you sign our CLA? Also, the test seems to be failing.

ebati:
I signed the CLA.

I removed the current version of binary parsing which lead to failing test (which used go version 1.17). For now i just updated CI to use go version 1.18.

knqyf263 (approved):
Amazing!

knqyf263:
Thank you so much! Looks perfect!

PR #90, opened 2022-04-02
feat(jar): Allow setting Maven Central URL using environment variable

kobus-v-schoor:
This should allow setting the URL which go-dep-parser uses to contact Maven Central when parsing JAR dependencies. It's an attempt to provide a solution for https://github.com/aquasecurity/trivy/issues/1869. I haven't developed in go before so not sure if this is a bad attempt or not :sweat_smile:


knqyf263 (approved):
LGTM👍  Thanks!

PR #91, opened 2022-04-05
Fix merge multiple pom imports

jkroepke:
In maven, a definition of a dependency will not override by any other definition.

```
% mvn -s testdata/conf/settings.xml -f testdata/import-dependency-management-multiple dependency:tree  | grep example-api
[ERROR] Failed to execute goal on project import: Could not resolve dependencies for project com.example:import:pom:2.0.0: org.example:example-api:jar:1.7.30 was not found in https://repo.maven.apache.org/maven2 during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
```

Previously, go-dep-parser detects version 1.1.1, since its defined in the second import bom.

See also: https://github.com/aquasecurity/trivy/issues/1943

### DIFF
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -267,7 +267,7 @@ func (p parser) dependencyManagement(deps []pomDependency, props properties) map
 			art := newArtifact(d.GroupID, d.ArtifactID, d.Version, props)
 			result, err := p.resolve(art)
 			if err == nil {
-				depManagement = p.mergeDependencyManagement(depManagement, result.dependencyManagement)
+				depManagement = p.mergeDependencyManagement(result.dependencyManagement, depManagement)
 			}
 			continue
 		}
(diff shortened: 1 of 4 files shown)


knqyf263 (approved):

knqyf263:
Great contribution! Thanks!!

knqyf263:
To fix the issue in Trivy, you need to

1. Update [fanal](https://github.com/aquasecurity/fanal) pointing to the new go-dep-parser
2. Update [trivy](https://github.com/aquasecurity/trivy) pointing to the updated fanal

jkroepke:
Hi @knqyf263 

thanks for the instructions. Is there are `go mod` command to archive this on fanal? Close to zero experience in golang here.

jkroepke:
Thanks. https://github.com/aquasecurity/fanal/pull/457

PR #92, opened 2022-04-07
feat(npm): add dependency relationships

AndreyLevchenko:
## Description
Pull Request introduces data structures to store parent dependencies for libraries. It should provide enough info to build dependency graph. Reference implementation for `package-lock.json` added


pkg/types/types.go
```diff
@@ -1,7 +1,25 @@
...
+
+func NewLibrary(name, version, license string) Library {
+	// TODO replace naive implementation
+	return Library{
+		Id:      fmt.Sprintf("%s@%s", name, version),
+		Name:    name,
+		Version: version,
+		License: license,
+	}
+}
```
knqyf263, line 25: What if we just define `ID()`?

```
func (lib Library) ID() string {
	return fmt.Sprintf("%s@%s", name, version)
}
```

pkg/types/types.go
```diff
@@ -1,7 +1,25 @@
...
+
 type Library struct {
+	Id      string
 	Name    string
 	Version string
 	License string `json:",omitempty"`
 }
+
+type Dependency struct {
+	Id        string
```
knqyf263, line 13: nit: I prefer ID.

```suggestion
	ID        string
```

AndreyLevchenko: I thought it is camelcase in `go`.

pkg/types/types.go
AndreyLevchenko: This is good question. Let's keep it as option. Once everything else is implemented we can return back and review it again

pkg/types/types.go
knqyf263: You can see here
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/types/types.go
knqyf263: It forces to add `types.NewLibrary()` everywhere and it makes review hard. It would be fine if it makes sense, but this ID is kind of like another format of `String()` because `Library` holds `name` and `version`. It looks like just one of expressions of `Library`. But yes, you can change it later.

pkg/nodejs/npm/parse.go
```diff
@@ -15,38 +16,72 @@ type Dependency struct {
...
 	Dev          bool
 	Dependencies map[string]Dependency
+	Requires     map[string]string
 }
 
-func Parse(r io.Reader) ([]types.Library, error) {
+func ID(name, version string) string {
+	// TODO replace naive implementation
+	return fmt.Sprintf("%s@%s", name, version)
+}
```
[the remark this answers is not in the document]
AndreyLevchenko: I'm going to make it part of interface (Since different parser can define it differently)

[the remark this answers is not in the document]
knqyf263: Sounds nice!

pkg/nodejs/yarn/parse.go
```diff
@@ -88,7 +88,7 @@ func Parse(r io.Reader) (libs []types.Library, err error) {
 			}
 
 			lib.Version = version
-			libs = append(libs, lib)
+			libs = append(libs, types.Library{Name: lib.Name, Version: lib.Version})
```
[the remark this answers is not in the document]
AndreyLevchenko: fixed

pkg/nuget/lock/parse.go
```diff
@@ -37,10 +37,7 @@ func Parse(r io.Reader) ([]types.Library, error) {
 				continue
 			}
 
-			lib := types.Library{
-				Name:    packageName,
-				Version: packageContent.Resolved,
-			}
+			lib := types.Library{Name: packageName, Version: packageContent.Resolved}
```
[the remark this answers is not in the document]
AndreyLevchenko: fixed

knqyf263 (approved):

PR #100, opened 2022-05-03
Support pom modules

ShiraCohen33:


knqyf263:
I'd recommend running tests locally. This change breaks many things.

PR #101, opened 2022-05-22
Calculate indirect libraries

ShiraCohen33:
Update the Indirect field if it's not in the Package dependencies from lock.file

### DIFF
--- a/pkg/nodejs/npm/parse.go
+++ b/pkg/nodejs/npm/parse.go
@@ -51,14 +51,6 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	return unique(libs), uniqueDeps(deps), nil
 }
 
-func appendDirectLibs(libs []types.Library, pkgs map[string]Dependency) {
-	for _, lib := range libs {
-		if _, ok := pkgs[lib.Name]; !ok {
-			lib.Indirect = true
-		}
-	}
-}
-
 func (p *Parser) parse(dependencies map[string]Dependency, versions map[string]string) ([]types.Library, []types.Dependency) {
 	// Update package name and version mapping.
 	for pkgName, dep := range dependencies {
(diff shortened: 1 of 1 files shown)


knqyf263 (changes requested):
Please try to add the metadata here since `Package` has the `Indirect` field.
https://github.com/aquasecurity/go-dep-parser/blob/66806cdbea8b5afaa484c84f0cc769ecd8b29154/pkg/nodejs/npm/parse.go#L68-L73

pkg/nodejs/npm/parse.go
```diff
@@ -128,3 +136,11 @@ func uniqueDeps(deps []types.Dependency) []types.Dependency {
 	}
 	return uniqDeps
 }
+
+func isIndirectLib(libName string, pkgs map[string]string) bool {
+	if _, ok := pkgs[libName]; !ok {
+		return true
+	}
+
+	return false
```
knqyf263, line 145:
```suggestion
	_, ok := pkgs[libName]
	return !ok
```

pkg/nodejs/npm/parse.go
```diff
@@ -43,12 +50,12 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 		return nil, nil, xerrors.Errorf("decode error: %w", err)
 	}
 
-	libs, deps := p.parse(lockFile.Dependencies, map[string]string{})
+	libs, deps := p.parse(lockFile.Dependencies, lockFile.Packages[""].Dependencies, map[string]string{})
```
knqyf263, line 53: for readability

```suggestion
	directDeps := lockFile.Packages[""].Dependencies
	libs, deps := p.parse(lockFile.Dependencies, directDeps, map[string]string{})
```

pkg/nodejs/npm/parse.go
```diff
@@ -43,12 +50,12 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
 
 	return unique(libs), uniqueDeps(deps), nil
 }
 
-func (p *Parser) parse(dependencies map[string]Dependency, versions map[string]string) ([]types.Library, []types.Dependency) {
+func (p *Parser) parse(dependencies map[string]Dependency, packages map[string]string, versions map[string]string) ([]types.Library, []types.Dependency) {
```
knqyf263, line 58:
```suggestion
func (p *Parser) parse(dependencies map[string]Dependency, directDeps map[string]string, versions map[string]string) ([]types.Library, []types.Dependency) {
```

knqyf263 (changes requested):

knqyf263 (approved):

Issue #102, opened 2022-05-31
feat(dotnet): Add support for parsing .deps.json files for .Net core


astevenson-microsoft:
.Net core generates a [appname].deps.json alongside the executable which contains the dependencies of that executable. Parsing it would provide information on dependencies for deployed .Net core executables.

The format is documented here: https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/specs/runtime-configuration-file.md

PR #103, opened 2022-05-31
feat(dotnet): Add support for parsing .Net core .deps.json files

astevenson-microsoft:
Add support for dotnet core .deps.json files

Fixes https://github.com/aquasecurity/go-dep-parser/issues/102


knqyf263 (approved):
Thanks for your great contribution!

PR #105, opened 2022-06-07
fix(pom): added check for looping for variable evaluation

DmitriyLewen:
## Description
added check for looping for variable evaluation

## Related issues
- https://github.com/aquasecurity/trivy/issues/2202

### DIFF
--- a/pkg/java/pom/artifact.go
+++ b/pkg/java/pom/artifact.go
@@ -5,6 +5,9 @@ import (
 	"os"
 	"regexp"
 	"strings"
+
+	"github.com/aquasecurity/go-dep-parser/pkg/log"
+	"golang.org/x/exp/slices"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/java/pom/artifact.go
```diff
@@ -105,9 +108,28 @@ func evaluateVariable(s string, props map[string]string) string {
...
 	}
 	return s
 }
+
+func printLoopedPropertiesStack(env string, usedProps []string) {
+	var s string
+	for _, prop := range usedProps {
+		s += fmt.Sprintf("%s -> ", prop)
+	}
+	log.Logger.Debugf("Unable to get variable. Lopped Properties was find: %s%s", s, env)
```
knqyf263, line 134: I think it doesn't happen usually. Can be a warning.

```suggestion
	log.Logger.Warnf("Lopped properties were detected: %s%s", s, env)
```

DmitriyLewen: Agree with you, thanks!

knqyf263 (approved):

PR #106, opened 2022-06-12
feat: add pnpm parser

### LINKED ISSUE #65: Support pnpm-lock.yaml

mycodeself:
Hey, this PR is ~~a draft~~ for adding a pnpm parser. I know that there is other open but there seems to be no progress so I wanted to offer my solution. I hope it doesn't matter.

The issue is #65 

What I think is missing:

- ~~Add detection of direct / indirect dependencies~~
- ~~Add more tests~~

Doubts I have:

- Should the dev vulnerabilities be ignored?

If you think there is anything else I should consider, let me know. I want to try to finish it this week.

### DIFF
--- a/pkg/nodejs/pnpm/parse.go
+++ b/pkg/nodejs/pnpm/parse.go
@@ -2,6 +2,7 @@ package pnpm
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 
 	"golang.org/x/xerrors"
@@ -57,7 +58,7 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 
 	libs, deps := p.parse(&lockFile)
 
-	return libs, deps, nil
+	return uniqueLibs(libs), uniqueDeps(deps), nil
 }
 
 func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency) {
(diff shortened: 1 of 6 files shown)


mycodeself:
Hi, I have added more tests and added the detection of direct / indirect dependencies. 

Let me know what you think!

knqyf263:
@mycodeself Thanks for your contribtuion!
@DmitriyLewen Could you take a look?

DmitriyLewen:
@knqyf263  Looks good!
About `dev Dependencies`: I agree with @mycodeself - i think we don't need to show dev deps. 
Also i thought about `optional Dependencies`: if i understand correctly - these packages may not be installed and we will get a false positive.

mycodeself:
@DmitriyLewen from [npm docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies) I understand that the `optionalDependencies` will be installed by default or at least attempt to install.

In fact, I have done some tests and as stated in the docs only when installing the dependencies with the `--no-optional` flag or when there is an error in fetching the dependency, the dependency is not installed.

Looking at the current parsers for nodejs:
- The `npm` parser will add this dependencies as currently is not filtering the optional dependencies and those are added in the `dependencies` object in the `package-lock.json`
- The `packagejson` parser will ignore the `optionalDependencies` as they are not added in the `dependencies` object in the `package.json` file
- The `yarn` parser will add this dependencies as they are added to the `yarn.lock` file

Therefore, if I am not mistaken, the optional dependencies are ignored only by the `packagejson`.

pkg/nodejs/pnpm/parse.go
```diff
@@ -0,0 +1,136 @@
...
+	var lockFile LockFile
+	decoder := yaml.NewDecoder(r)
+	err := decoder.Decode(&lockFile)
+	if err != nil {
+		return nil, nil, xerrors.Errorf("decode error: %w", err)
+	}
+
+	libs, deps := p.parse(&lockFile)
+
+	return uniqueLibs(libs), uniqueDeps(deps), nil
```
knqyf263, line 61: Do we need deduplication here? IIUC, we iterate `packages` that are unique.

mycodeself: I have done some checking and you are correct! I have removed the unnecessary deduplication

let me know if is ok now :)

knqyf263 (approved):

knqyf263:
Great work! Thanks for your contribution!

knqyf263:
@mycodeself We recently deprecated fanal. Could you open a PR in Trivy?

mycodeself:
@knqyf263 I'll try to take a look when I have some free time !

PR #109, opened 2022-06-20
fix(bundler): drop platform from version string

### LINKED ISSUE #108: Bundler uses different RegExp for parsing lockfile lines
I've been troubleshooting https://github.com/aquasecurity/go-gem-version/issues/1 and realized this might be due to a misalignment on how this library parses dependencies, versus how bundler does it.

You can see the regex bundler uses to parse a dependency in lockfiles [here](https://github.com/rubygems/rubygems/blob/75bd5fec39bc9ba53bce07c5aae47cf0adaf4e81/bundler/lib/bundler/lockfile_parser.rb#L145-L153).

This is the important piece:
```
      (?:#{space}\(([^-]*)                               # Space, followed by version
      (?:-(.*))?\))?                                     # Optional platform
```
Bundler assumes there will be no `-` in the version string and it parses the platform separately (e.g. `-x86_64-darwin`)

So for `    nokogiri (1.13.1-x86_64-darwin)`:
- **bundler**: name=`nokogiri` version=`1.13.1` platform=`x86_64-darwin`
- **go-dep-parser**: name=`nokogiri` version=`1.13.1-x86_64-darwin`

When `go-gem-version` sees `1.13.1-x86_64-darwin`, it will run into https://github.com/aquasecurity/go-gem-version/issues/1 (and this causes Trivy to skip checks on dependencies that have platformed versions.)

I'll try work on a PR but if someone else wants to take a stab at it, please don't hesitate.

kgeorgiou:
Fixes #108 

Currently when a dependency has a platform appended to its version, this library will keep the platform string in the parsed version.

**Example:** 
`tailwindcss-rails (2.0.3-x86_64-darwin)` => `Name: tailwindcss-rails`, `Version: 2.0.3-x86_64-darwin`

This causes problems downstream to [go-gem-version](https://github.com/aquasecurity/go-gem-version/blob/main/version.go#L42) and as a result to Trivy, which will skip checks on any dependencies with platform appended versions.

👉 This PR drops the platform from the version string.

**Example:**
`tailwindcss-rails (2.0.3-x86_64-darwin)` => `Name: tailwindcss-rails`, `Version: 2.0.3`

I based this change on how bundler is parsing dependencies in a lockfile. It assumes the first `-` separates the version from the optional platform string:
https://github.com/rubygems/rubygems/blob/4b1b2b7d457ae664a1f99001a70b10ff5993f935/bundler/lib/bundler/lockfile_parser.rb#L145-L153


knqyf263 (approved):
Awesome!

knqyf263:
@kgeorgiou We've done. Could you bump go-dep-parser in Trivy?
https://github.com/aquasecurity/trivy/pull/2356

PR #115, opened 2022-07-14
fix(gemspec): split licenses by ', '

DmitriyLewen:
## Description
There are licenses like `"Apache 2.0".freeze`. 
We incorrectly split this license as `Apache` and `2.0.freeze`.

Changed separator to `,`.

### DIFF
--- a/pkg/ruby/gemspec/parse.go
+++ b/pkg/ruby/gemspec/parse.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
-	"unicode"
 
 	"golang.org/x/xerrors"
 
@@ -39,7 +38,7 @@ var (
 	// Capture the value of "licenses"
 	// e.g. s.license = ["MIT".freeze, "BSDL".freeze]
 	//      => "MIT".freeze, "BSDL".freeze
-	licensesRegexp = regexp.MustCompile(`\.licenses\s*=\s*\[(?P<licenses>.+)\]`)
+	licensesRegexp = regexp.MustCompile(`\.licenses\s*=\s*\[(?P<licenses>.+)]`)
 )
 
 type Parser struct{}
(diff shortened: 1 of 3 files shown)


pkg/ruby/gemspec/parse.go
```diff
@@ -121,9 +120,7 @@ func trim(s string) string {
 func parseLicenses(s string) string {
 	// e.g. `"Ruby".freeze, "BSDL".freeze`
 	//      => {"\"Ruby\".freeze", "\"BSDL\".freeze"}
-	ss := strings.FieldsFunc(s, func(r rune) bool {
-		return unicode.IsSpace(r) || r == ','
-	})
+	ss := strings.Split(s, ", ")
```
knqyf263, line 123: What about "GPL,Apache"?

DmitriyLewen: Oh... yes, your are right.
Fixed.

pkg/ruby/gemspec/parse.go
```diff
@@ -39,7 +38,7 @@ var (
 	// Capture the value of "licenses"
 	// e.g. s.license = ["MIT".freeze, "BSDL".freeze]
 	//      => "MIT".freeze, "BSDL".freeze
-	licensesRegexp = regexp.MustCompile(`\.licenses\s*=\s*\[(?P<licenses>.+)\]`)
```
knqyf263, line 42: Do we need to remove this `\`?

DmitriyLewen: This `\` is redundant, because regex doesn't have character set opener (`[`).
Perhaps this can be confuse.
Do you want me to return `/`?

knqyf263: > regex doesn't have character set opener ([).

`[]` has a special meaning in regexp, so I thought we needed to escape them like `\[\]`.
https://github.com/google/re2/wiki/Syntax

Actually, an opener is still escaped `\[`. Is that escape also unneeded?
`\.licenses\s*=\s*[(?P<licenses>.+)]`

DmitriyLewen: My IDE(`goland`) recommended me to remove redundant symbols.
I checked it with [regex101](https://regex101.com/r/rCuT62/1). It really works.
But I didn't find any explanation why it works.

Looks like there are 2 cases:
- regex contain `[`. Then `]` will complete character set(e.g. [[cba]](https://regex101.com/r/CjZXNs/2))
- regex doesn't contain `[` (or this character is escaped). Then `]` will be common character(e.g. [cba\]](https://regex101.com/r/603dQJ/1) or [\\\[cba\]](https://regex101.com/r/603dQJ/2))

knqyf263: Interesting. Yes, we should escape `[` and `]` both for clarity.

knqyf263 (approved):

PR #116, opened 2022-07-15
fix(pom): change parent priority 

DmitriyLewen:
## Description
When there is parent of based `pom.xml` and parent of child dependency, Maven takes the depmanager from the parent of the base `pom.xml`.

Diagram to clarify:

dia1

## Related issues:
- aquasecurity/trivy/issues/2430


pkg/java/pom/parse.go
```diff
@@ -291,17 +296,17 @@ func (p parser) mergeDependencyManagement(a, b map[string]pomDependency) map[str
 	return a
 }
 
-func (p parser) parseDependencies(deps []pomDependency, props map[string]string, depManagement map[string]pomDependency,
+func (p parser) parseDependencies(deps *pom, props map[string]string, depManagement map[string]pomDependency,
 	exclusions map[string]struct{}) []artifact {
 	var dependencies []artifact
-	for _, d := range deps {
+	for _, d := range deps.content.Dependencies.Dependency {
```
knqyf263, line 302: Could you clarify the reason for this change?

pkg/java/pom/parse.go
```diff
@@ -247,8 +247,13 @@ func (p *parser) analyze(pom *pom, exclusions map[string]struct{}) (analysisResu
 	depManagement := p.dependencyManagement(pom.content.DependencyManagement.Dependencies.Dependency, props)
 	depManagement = p.mergeDependencyManagement(parent.dependencyManagement, depManagement)
 
+	// depManagement deps from root file have higher priority
+	for k, v := range rootDependencyManagement {
+		depManagement[k] = v
+	}
+
```
knqyf263, line 254: Can we put it into `mergeDependencyManagement`?

```
func (p parser) mergeDependencyManagements(depManagements ...map[string]pomDependency) map[string]pomDependency {
	new := map[string]pomDependency{}
	// The preceding argument takes precedence.
	for i := len(depManagements) - 1; i >= 0; i-- {
		for key, value := range depManagements[i] {
			new[key] = value
		}
	}
	return new
}
```

pkg/java/pom/testdata/repository/org/example/example-child-parent/2.0.0/example-child-parent-2.0.0.pom
```diff
@@ -0,0 +1,23 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
knqyf263, line 1: I took a read at the doc again.
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management

I found the following statement.

>Finally, since d is specified in B's dependency management section, should d be a dependency (or transitive dependency) of a or c, version 1.0 will be chosen - again because dependency management takes precedence over dependency mediation and also because the current POM's declaration takes precedence over its parent's declaration.

Another question came to my mind. You added 4 poms for testing. If all of them have `dependencyManagement`, what is the priority on `org.example:example-child:2.0.0`? According to the above statement, `dependencyManagement` in `org.example:example-child:2.0.0` is the highest priority. What about others?

1. org.example:example-child:2.0.0 (repository/org/example/example-child/2.0.0/example-child-2.0.0.pom)
2. com.example:parent:3.0.0 (transitive-parents/pom.xml)
3. com.example:base:4.0.0 (transitive-parents/base/pom.xml)
4. org.example:example-child-parent:2.0.0 (repository/org/example/example-child/2.0.0/example-child-parent-2.0.0.pom)

or

1. org.example:example-child:2.0.0 (repository/org/example/example-child/2.0.0/example-child-2.0.0.pom)
2. com.example:base:4.0.0 (transitive-parents/base/pom.xml)
3. com.example:parent:3.0.0 (transitive-parents/pom.xml)
4. org.example:example-child-parent:2.0.0 (repository/org/example/example-child/2.0.0/example-child-parent-2.0.0.pom)

or

1. org.example:example-child:2.0.0 (repository/org/example/example-child/2.0.0/example-child-2.0.0.pom)
2. com.example:parent:3.0.0 (transitive-parents/pom.xml)
3. org.example:example-child-parent:2.0.0 (repository/org/example/example-child/2.0.0/example-child-parent-2.0.0.pom)
4. com.example:base:4.0.0 (transitive-parents/base/pom.xml)

or something else? In your current example, only parents (`com.example:parent:3.0.0` and `org.example:example-child-parent:2.0.0` ) have `dependencyManagement`, but what if children also have `dependencyManagement`?

knqyf263: To clarify my question, I drew a diagram. The version of `jackson-databind` should be `3.0.0`, but what is next if `3.0.0` is not specified? `1.0.0` or `2.0.0`?

image

pkg/java/pom/parse.go
DmitriyLewen: you are right. It is extra changes. I revert it.

pkg/java/pom/parse.go
DmitriyLewen: Good idea!
Done

pkg/java/pom/testdata/repository/org/example/example-child-parent/2.0.0/example-child-parent-2.0.0.pom
DmitriyLewen: > According to the above statement, dependencyManagement in org.example:example-child:2.0.0 is the highest priority.


It has no logic. We should be able to change any child dependency in the base pom.
Perhaps you mixed up `example-child` and `base`? `project B` from maven docs is `base` in your example: `When maven is run on project B`. Or i missed something?

I reproduced your example.
Maven has the following priority:
1. com.example:base:4.0.0
2. com.example:parent:3.0.0
3. org.example:example-child:2.0.0
4. org.example:example-child-parent:2.0.0

also i checked Trivy with this PR. Trivi works the same way.

knqyf263: Hmm. Even though `org.example:example-child:2.0.0` has `dependencyManagement` and `dependencies` both, Does `com.example:base:4.0.0` take precedence? In the above example, is `2.0.0` selected rather than `3.0.0`?

```
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>example</groupId>
      <artifactId>jackson-databind</artifactId>
      <version>3.0.0</version>
    </dependency>
  </dependencies>
</dependencyManagement>

<dependencies>
  <dependency>
    <groupId>example</groupId>
    <artifactId>jackson-databind</artifactId>
  </dependency>
</dependencies>
```

DmitriyLewen: `dependencyManagement` from root `pom.xml` has highest priority.
e.g. we have:
`com.example:base:4.0.0`:
```
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>example</groupId>
      <artifactId>jackson-databind</artifactId>
      <version>2.2.2</version>
    </dependency>
  </dependencies>
</dependencyManagement>

<dependencies>
  <dependency>
    <groupId>org.example</groupId>
    <artifactId>example-child</artifactId>
    <version>2.0.0</version>
  </dependency>
</dependencies>
```

and `org.example:example-child:2.0.0`:
```
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>example</groupId>
      <artifactId>jackson-databind</artifactId>
      <version>3.3.3</version>
    </dependency>
  </dependencies>
</dependencyManagement>

<dependencies>
  <dependency>
    <groupId>example</groupId>
    <artifactId>jackson-databind</artifactId>
  </dependency>
</dependencies>
```

In this case version of `jackson-databind` is `2.2.2`.

knqyf263: Got it. Thanks for clarifying.

pkg/java/pom/parse.go
```diff
@@ -245,7 +245,7 @@ func (p *parser) analyze(pom *pom, exclusions map[string]struct{}) (analysisResu
 
 	// Extract and merge dependencies under "dependencyManagement"
 	depManagement := p.dependencyManagement(pom.content.DependencyManagement.Dependencies.Dependency, props)
-	depManagement = p.mergeDependencyManagement(parent.dependencyManagement, depManagement)
+	depManagement = p.mergeDependencyManagements(rootDependencyManagement, depManagement, parent.dependencyManagement)
```
knqyf263, line 248: Can you please add a comment about the precedence order?

pkg/java/pom/parse.go
```diff
@@ -223,7 +223,7 @@ type analysisResult struct {
 	modules              []string
 }
 
-func (p *parser) analyze(pom *pom, exclusions map[string]struct{}) (analysisResult, error) {
+func (p *parser) analyze(pom *pom, exclusions map[string]struct{}, rootDependencyManagement map[string]pomDependency) (analysisResult, error) {
```
knqyf263, line 226: It doesn't have to be root, right? It just comes from dependents.
- package A
  - package B
    - package C

If I understand correctly, in the above case, package C is affected by `dependencyManagement` in package A and package B. Any other better name? `rootDependencyManagement` sounds only from package A. Or am I wrong?

DmitriyLewen: you are right.
But I couldn't think of a better name.

Maybe `rootPomsDependencyManagement` or `rootPackagesDependencyManagement`?
Or perhaps you have some ideas?

pkg/java/pom/parse.go
DmitriyLewen: Done

pkg/java/pom/parse.go
knqyf263: I came up with `dependentDependencyManagement`, but it could be confusing.

knqyf263 (approved):

PR #121, opened 2022-08-08
Parse deps specified as PackageReference in .csproj files

Sjord:
C# project files, with the extension .csproj, are XML files that can specify project dependencies in `<PackageReference>` tags.

See also:
* https://github.com/aquasecurity/trivy/issues/2668
* https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files


Sjord:
From the CLA:

> You represent that each of Your Contributions is Your original creation

Well, I copied pkg/nuget/config to pkg/nuget/csproj and modified things to get it to work on .csproj files. I guess that's OK, right?

knqyf263:
>Well, I copied pkg/nuget/config to pkg/nuget/csproj and modified things to get it to work on .csproj files. I guess that's OK, right?

Yes, thanks for signing.

@DmitriyLewen Could you review this PR?

pkg/nuget/csproj/parse.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var cfgData config
+	if err := xml.NewDecoder(r).Decode(&cfgData); err != nil {
+		return nil, nil, xerrors.Errorf("failed to decode .csproj file: %w", err)
+	}
+
+	libs := make([]types.Library, 0)
+	for _, pkg := range cfgData.Packages {
+		if pkg.Include == "" || pkg.PrivateAssets != "" {
```
DmitriyLewen, line 39: looks like we need use `pkg.PrivateAssets != "all"`here
if i understand correctly: this and other asset tags work with [dependency parts](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets).
I think we need skip dependency only if all parts are private.

Also i am not sure about `pkg.ExcludeAssets == "all"`.

pkg/nuget/csproj/testdata/dev_dependency.csproj
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <TargetFramework>net46</TargetFramework>
```
DmitriyLewen, line 4: @Sjord can you help me understand - can we use `TargetFramework` value for condition? [condition](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#adding-a-packagereference-condition)?

Sjord: No. Realistically, a [csproj that has a condition on `TargetFramework`](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Newtonsoft.Json.csproj) would also define multiple supported target frameworks. The exact target framework that is currently being built is one of those, but which one is not specified.

DmitriyLewen: I see. Thanks! I understand now.

pkg/nuget/csproj/parse.go
Sjord: > looks like we need use pkg.PrivateAssets != "all" here

Yes, that seems indeed correct.

PrivateAssets does not precisely correspond to devDependencies, but in practice `PrivateAssets = all` seems to be used for development dependencies.

I also discovered that PrivateAssets could be an attribute on PackageReference instead of a tag, and that it is case insensitive. I'll change the code to case-insensensitively compare either the tag or the attribute to "all".

Sjord: As for ExcludeAssets, there are two common options:

* `ExcludeAssets=all`, this basically means that the dependency isn't a dependency. Nothing is loaded from it. I am not sure why you would want this, but I think we should exclude it.
* `ExcludeAssets=runtime`, this means that the assemblies are not available on runtime, but only during compile time. It's used for analyzers, code generators, etc. So it seems to indicate a development dependency, and we should exclude it.

Sjord:
I now put this in the `nuget` directory, but I later saw that there is also a `dotnet` directory. Is `csproj` correct in the `nuget` dir, or does `dotnet` make more sense?

pkg/nuget/csproj/parse.go
```diff
@@ -0,0 +1,52 @@
...
+			continue
+		}
+
+		lib := types.Library{
+			Name:    pkg.Include,
+			Version: pkg.Version,
```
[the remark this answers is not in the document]
Sjord: I [changed the code](https://github.com/aquasecurity/go-dep-parser/pull/121/commits/2e57fbafe8d43d821a3a94a3d7f8aeb2c3877f95) to trim `.*`. I don't think it is elegant to reduce the version like this, but I guess it's the best option that's possible. It results in desired behavior in trivy, because if 6.0.4.5 is vulnerable, you don't want to specify 6.0.4.* as dependency.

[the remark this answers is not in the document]
DmitriyLewen: > because if 6.0.4.5 is vulnerable, you don't want to specify 6.0.4.* as dependency.
Agree with you!

DmitriyLewen:
LGTM

Sjord:
@knqyf263 This has been reviewed, could you merge this?

knqyf263:
The result can be incomplete since it extracts only packages having exact versions, right? We're currently discussing whether we should display these incomplete results or encourage users to generate a lock file. If we show any result, it looks like we properly determined installed packages, but it is not the case actually. It can confuse users.
Any thoughts? Please correct me if my above assumption is wrong.

Sjord:
Yes, this is an opportunistic approach where we can verify some dependencies, but in some cases determining the exact version is not possible. I think this is preferable to not scanning dependencies at all, but I agree this may give false confidence. If a lock file is present, that should be scanned instead. If no lock file is present, it would be a good idea to report that to the user.

As a security consultant, I use trivy to scan my client's projects. Finding any vulnerable dependency allows me to point the client to the importance of keeping packages up to date, and advice them to use a lock file. So completeness is not very important to me, as I use this to indicate a general problem instead of finding which packages need updating.

bjuraga:
Any chance this gets merged? As it is, it does have some limitations, but i would split the implementation to 2 phases, Optimistic and Deterministic. At the moment we have nothing in the SBOM, and with this implementation, we can at least get most of the package names.

afrischk:
Any update on this?

bjuraga:
> Any update on this?

Might be better to use the *.deps.json from your bin folder. We have had a great success in doing so, and that file contains the actual dependencies instead of the requested ones.

afrischk:
This holds true for .NET Core and .NET 5+. I was hoping to use Trivy on an older .NET project with no .deps.json available but PackageReference in .csproj files.

afrischk:
Ok, I found a solution. After reading the comment from @knqyf263 carefully I noticed 'encourage users to generate a lock file'. I read through microsofts docs about `packages.lock.json` and added a `Directory.Build.props` file with the following contents to the root directory of my solution.
```
<Project>
 <PropertyGroup>
   <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
 </PropertyGroup>
</Project>
```
All subprojects are now generating a `packages.lock.json` and Trivy happily parses them :-) I hope this is helpful for anybody who came across this issue.

itrf98:
Dear Team,

Any news about this ? Are you planning to merge it ?

Thank you in advance

afrischk:
@itrf98 have you read @bjuraga , @knqyf263 and my comment about the `json` files? Is this a solution for you?

PR #122, opened 2022-08-08
feat(Rust): Extract dependency graph from Rust binaries

tofay:
Following on from https://github.com/rust-secure-code/cargo-auditable/issues/61, this updates go-dep-parser to extract the dependency graph.

I thought it made sense for the root package to simply be the only direct package, instead of the root and its direct dependencies being marked as direct.


pkg/utils/utils.go
```diff
@@ -41,3 +41,7 @@ func MergeMaps(parent, child map[string]string) map[string]string {
 	}
 	return parent
 }
+
+func PackageID(name, version string) string {
```
knqyf263, line 45: Just FYI: We wanted to change how to generate package IDs depending on ecosystem. For example, Java will probably create an ID like `com.google.code.gson:gson`. That's why we kept `ID()` in each package. But we can refactor it later. Thanks.

knqyf263 (approved):

knqyf263:
Great! Thanks a lot!

PR #123, opened 2022-08-10
feat(pom): add dependency tree

Sq34sy:
I added a dependency graph when parsing pom files. This is the pull request for issue https://github.com/aquasecurity/go-dep-parser/issues/126

I used https://github.com/aquasecurity/go-dep-parser/pull/116 as a starting point and changed some tests.

In the following exmaple i will use the pom.xml file from https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/java/pom/testdata/happy/pom.xml .
Before the changes the output might have looked something like this:

```json
{
  "libs": [
    {
      "Name": "com.example:happy",
      "Version": "1.0.0"
    },
    {
      "Name": "org.example:example-api",
      "Version": "1.7.30"
    }
  ],
  "deps": null
}
```

After the changes it will look something like this:

```json
{
  "libs": [
    {
      "ID": "com.example.happy:1.0.0",
      "Name": "com.example:happy",
      "Version": "1.0.0"
    },
    {
      "ID": "org.example.example-api:1.7.30",
      "Name": "org.example:example-api",
      "Version": "1.7.30"
    },
    {
      "ID": "org.example.utils:1.7.30",
      "Name": "org.example:utils",
      "Version": "1.7.30"
    }
  ],
  "deps": [
    {
      "ID": "com.example.happy:1.0.0",
      "DependsOn": [
        "org.example.example-api:1.7.30"
      ]
    },
    {
      "ID": "org.example.example-api:1.7.30",
      "DependsOn": [
        "org.example.utils:1.7.30"
      ]
    }
  ]
}
```

All dependecies, no matter if they are found directly or indirectly, will be listed as `libs`. This the original version is preffered the i can alter the changes.

### DIFF
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -170,8 +170,8 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency, 
 		if !art.IsEmpty() {
 			// Override the version
 			uniqArtifacts[art.Name()] = artifact{
-				Version: art.Version,
-				License: art.License,
+				Version:  art.Version,
+				Licenses: art.Licenses,
 			}
 		}
 	}
@@ -181,7 +181,7 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency, 
 		libs = append(libs, types.Library{
 			Name:    name,
 			Version: art.Version.String(),
-			License: art.License,
+			License: art.JoinLicenses(),
 		})
 	}
 
@@ -340,7 +340,7 @@ func (p *parser) resolveDepManagement(props map[string]string, depManagement []p
 	// Managed dependencies with a scope of "import" should be processed after other managed dependencies.
 	// cf. https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#importing-dependencies
 	for _, imp := range imports {
-		art := newArtifact(imp.GroupID, imp.ArtifactID, imp.Version, "", props)
+		art := newArtifact(imp.GroupID, imp.ArtifactID, imp.Version, nil, props)
 		result, err := p.resolve(art, nil)
 		if err != nil {
 			continue
@@ -390,7 +390,7 @@ func excludeDep(exclusions map[string]struct{}, art artifact) bool {
 
 func (p *parser) parseParent(currentPath string, parent pomParent) (analysisResult, error) {
 	// Pass nil properties so that variables in <parent> are not evaluated.
-	target := newArtifact(parent.GroupId, parent.ArtifactId, parent.Version, "", nil)
+	target := newArtifact(parent.GroupId, parent.ArtifactId, parent.Version, nil, nil)
 	// if version is property (e.g. ${revision}) - we still need to parse this pom
 	if target.IsEmpty() && !isProperty(parent.Version) {
 		return analysisResult{}, nil
--- a/pkg/java/pom/artifact.go
+++ b/pkg/java/pom/artifact.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/aquasecurity/go-dep-parser/pkg/log"
+	"github.com/samber/lo"
 	"golang.org/x/exp/slices"
 )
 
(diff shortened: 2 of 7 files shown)


knqyf263:
Hi, thanks for your contribution! 

> All dependecies, no matter if they are found directly or indirectly, will be listed as libs

I think we need to determine if a package is directly used or not so that we can create a dependency graph. Am I missing something?

Sq34sy:
> I think we need to determine if a package is directly used or not so that we can create a dependency graph. Am I missing something?

When it comes to creating a dependency graph, it doesn't really matter what is in `libs`. It is possible to create a graph using only data from `deps`. Basically an element in the deps list represents a tree with the depth of 1. These trees can be unified to create a bigger graph. 

An Example: 

```json
{
  "deps": [
    {
      "ID": "Root",
      "DependsOn": [
        "A",
        "B"
      ]
    },
    {
      "ID": "B",
      "DependsOn": [
        "C",
        "D"
      ]
    }
  ]
}
```

<p>
  
   
</p>

image

Regardless of all that. I might be better if `libs` has only direct dependencies, i.e. only dependecies which are specified in the pom.xml file. I do not know enough about this project, therefore i dont want to make that decisions.

pkg/java/pom/parse.go
```diff
@@ -111,28 +111,58 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency,
...
 	uniqArtifacts := map[string]version{}
 
 	// Iterate direct and transitive dependencies
 	for !queue.IsEmpty() {
 		art := queue.dequeue()
 
+		{
+			id := id(art)
+			_, contains := processed[id]
+			keys := make([]string, 0, len(processed))
```
AndreyLevchenko, line 124: what is purpose of `keys` slice?

pkg/java/pom/parse.go
```diff
@@ -111,28 +111,54 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency,
...
 			}
-			libs = append(libs, moduleLibs...)
-			if deps != nil {
+
+			for _, lib := range moduleLibs {
+				_, contains := processed[lib.ID]
+				if contains {
+					continue
+				} else {
+					groupID, artifactID, verion, err := deconstructID(lib.ID)
```
AndreyLevchenko, line 144: ignored map value contains artifact struct which includes artifactID, groupID etc, so it seems `deconstructID()` is not required.

pkg/java/pom/parse.go
```diff
@@ -111,28 +111,54 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency,
...
 	var libs []types.Library
 	var deps []types.Dependency
+	processed := map[string]artifact{}
 	uniqArtifacts := map[string]version{}
 
 	// Iterate direct and transitive dependencies
 	for !queue.IsEmpty() {
 		art := queue.dequeue()
 
+		{
```
AndreyLevchenko, line 121: 1. `if` branch with `continue` is not covered by unit tests
2. could you use if with initialization statement
```go
		id := id(art)
		if _, contains := processed[id]; contains {
			continue
		} else {
			processed[id] = art
		}
```

pkg/java/pom/parse.go
```diff
@@ -164,16 +196,47 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency,
 	}
 
 	// Convert to []types.Library
-	for name, ver := range uniqArtifacts {
```
AndreyLevchenko, line 167: you are using `processed` instead of `uniqArtifacts` here. I would suggest replace  `uniqArtifacts` in other places as well.
Keeping both collection may cause confusion in the future

Tsuesun:
Hi @Sq34sy & @AndreyLevchenko,

Any chance we can get this PR merged in? I am actually using this to build a dep tree so I can find the root dependencies found by trivy

DmitriyLewen:
Hello @Sq34sy 
Thanks for your work!

Can you rebase this PR?

Regards, Dmitriy

Sq34sy:
Hi,

Sorry, I had a little mishap. I can rebase properly it in the following days. Should I open a new Pull Request or can this one somehow be reopen?

Regards

DmitriyLewen:
Hello @Sq34sy 

> can this one somehow be reopen?

I also can't reopen this PR.


I didn't understand what happened with your PR and created #243 😄 .
You can open new PR or review #243.

PR #127, opened 2022-08-19
feat(gradle): add *gradle.lockfile parser

DmitriyLewen:
## Description
Add [*gradle.lockfile](https://docs.gradle.org/current/userguide/dependency_locking.html#lock_state_location_and_format) parser.

## Related issue:
- aquasecurity/trivy/issues/2707

### DIFF
--- a/pkg/gradle/lockfile/parse.go
+++ b/pkg/gradle/lockfile/parse.go
@@ -1,0 +1,39 @@
+package lockfile
+
+import (
+	"bufio"
+	"strings"
+
+	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
+)
+
+type Parser struct{}
+
+func NewParser() types.Parser {
+	return &Parser{}
+}
+
+func (Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var libs []types.Library
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "#") { // skip comments
+			continue
+		}
+
+		// dependency format: group:artifact:version=classPaths
+		dep := strings.Split(line, ":")
+		if len(dep) != 3 { // skip the last line with lists of empty configurations
+			continue
+		}
+		libs = append(libs, types.Library{
+			Name:    strings.Join(dep[:2], ":"),
+			Version: strings.Split(dep[2], "=")[0], // remove classPaths
+		})
+
+	}
+	return utils.UniqueLibraries(libs), nil, nil
+}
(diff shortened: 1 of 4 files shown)


pkg/gradle/lockfile/parse.go
```diff
@@ -0,0 +1,39 @@
...
+func NewParser() types.Parser {
+	return &Parser{}
+}
+
+func (Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var libs []types.Library
+	scanner := bufio.NewScanner(r)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "#") { // skip comments
```
afdesk, line 23: I'd use something like [TrimSpace](https://pkg.go.dev/strings#TrimSpace) here, to skip next comments:
```yaml
 # comment
   # comment2
````

File: pkg/gradle/lockfile/parse.go [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+			continue
+		}
+
+		// dependency format: group:artifact:version=classPaths
+		dep := strings.Split(line, ":")
+		if len(dep) != 3 { // skip the last line with lists of empty configurations
```
afdesk, line 29: is this check enough? i mean that users can have strange (incorrect) lockfiles: https://github.com/tadfisher/gradle2nix/blob/master/gradle/dependency-locks/settings-classpath.lockfile

or it doesn't matter, right?

pkg/gradle/lockfile/testdata/happy.lockfile
```diff
@@ -0,0 +1,7 @@
+# This is a Gradle generated file for dependency locking.
+# Manual edits can break the build and are not advised.
+# This file is expected to be part of source control.
+cglib:cglib-nodep:2.1.2=testRuntimeClasspath,classpath
+org.springframework:spring-asm:3.1.3.RELEASE=classpath
```
afdesk, line 5: I'd also add a record from the gradle sample:
```
org.springframework:spring-beans:5.0.5.RELEASE=compileClasspath, runtimeClasspath
```

pkg/gradle/lockfile/parse.go
afdesk: or it doesn't matter here too...

afdesk (approved):

pkg/gradle/lockfile/parse.go
DmitriyLewen: dependencies from this file don't have `=classPaths`. The rest looks default.
I think we can stay this check. If  there are problems with this, we will add more checks.

pkg/gradle/lockfile/parse.go
DmitriyLewen: It is make sense.
I am not sure that is possible, but i added trim to be safe.

pkg/gradle/lockfile/testdata/happy.lockfile
DmitriyLewen: Done

PR #129, opened 2022-08-24
fix(rust): add skipped package IDs

afdesk:
There was skipped package IDs, so `trivy` can't match dependencies

### DIFF
--- a/pkg/rust/binary/parse.go
+++ b/pkg/rust/binary/parse.go
@@ -58,7 +58,7 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 			for _, dep_idx := range pkg.Dependencies {
 				dep := info.Packages[dep_idx]
 				if dep.Kind == rustaudit.Runtime {
-					childDeps = append(childDeps, pkgID)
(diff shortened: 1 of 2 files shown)


knqyf263:
Tests are failing.

afdesk:
> Tests are failing.

tests are the greatest thing, it found a bug

pkg/rust/binary/parse.go
```diff
@@ -46,7 +46,9 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	var deps []types.Dependency
 	for _, pkg := range info.Packages {
 		if pkg.Kind == rustaudit.Runtime {
+			pkgID := utils.PackageID(pkg.Name, pkg.Version)
 			libs = append(libs, types.Library{
+				ID:       pkgID,
```
knqyf263, line 51: Is the problem that `ID` is missing in `types.Library`?

afdesk: yes, ID is needed

afdesk: all packages have an empty string instead of ID now.

knqyf263: OK

knqyf263 (approved):

PR #132, opened 2022-09-07
Fix #131. Support revisions in Conan parser

### LINKED ISSUE #131: New Conan parser does not handle revisions properly
Hi,

I saw the new parser for `conan.lock` files was pushed a few hours ago, so I decided to give a try on some of my projects.
However, the parser fails to parse my sample file because I have enabled Conan's revision support (see https://docs.conan.io/en/latest/versioning/revisions.html).
In my case, the references inside the lock file look like `pkg/version#rrev` (rrev = recipe revision), e.g. `openssl/3.0.5#8743768a28916117e28629f0c4921a29`.

I end up with the following error when parsing the file:
`WARN     version error (3.0.5#8743768a28916117e28629f0c4921a29): malformed version: 3.0.5#8743768a28916117e28629f0c4921a29`

Unfortunately I cannot share my sample file, sorry.

I think the fix would be a single one-line change at https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/c/conan/parse.go#L97 to also split the string on "#" (in addition to "@") when looking for the package's version.

Regards,
François

fpoirotte:
This adds support for recipe revisions inside the Conan parser.

### DIFF
--- a/pkg/c/conan/parse.go
+++ b/pkg/c/conan/parse.go
@@ -91,10 +91,13 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 }
 
 func parseRef(ref string) (types.Library, error) {
-	// ref format examples: 'pkga/0.1@user/testing'
+	// full ref format: package/version@user/channel#rrev:package_id#prev
+	// various examples:
+	// 'pkga/0.1@user/testing'
 	// 'pkgb/0.1.0'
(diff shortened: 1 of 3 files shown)


knqyf263 (approved):

knqyf263:
Thanks. If you don't mind, it would be appreciated if you open a PR in Trivy as well to bump the go-dep-parser version.

PR #134, opened 2022-09-22
feat(npm): add dependencies location

DmitriyLewen:
## Description
Add dependency line numbers from `package-lock.json` files


pkg/utils/utils.go
```diff
@@ -20,16 +21,21 @@ func UniqueStrings(ss []string) []string {
...
+	}
+	unique := map[string]types.Library{}
 	for _, lib := range libs {
 		identifier := fmt.Sprintf("%s@%s", lib.Name, lib.Version)
-		if _, ok := unique[identifier]; !ok {
-			unique[identifier] = struct{}{}
-			uniqLibs = append(uniqLibs, lib)
+		if l, ok := unique[identifier]; !ok {
+			unique[identifier] = lib
+		} else if len(lib.Locations) > 0 {
```
knqyf263, line 32: Does it happen that the same dependency appears in multiple locations?

DmitriyLewen: Right.
example from `testdata/package-lock_many.json`:
1) [L148-154](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L148-L154)
2) [L486-L492](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L486-L492)
3) [L535-L541](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L535-L541)

knqyf263: I think we should take only the top-level location. Others are just representing the requirements of other dependencies.
https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L310-L317

DmitriyLewen: There is case where top-level dependencies doesn't contain requirements of other dependencies.
e.g. `debug`
Top-level deps only [contain](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L302-L323)  `v3.2.6`.
But [body-parser](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L131-L162) needs `v2.6.9`.
Also for [express](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L449-L500) and [finalhandler](https://github.com/aquasecurity/go-dep-parser/blob/d3a51fe400e451f6c2725db8eb1466c9e228530d/pkg/nodejs/npm/testdata/package-lock_many.json#L521-L549).
We [will include](https://github.com/aquasecurity/go-dep-parser/blob/a5451f44782082e09e05ff53d33ab6d70e9e6448/pkg/nodejs/npm/parse.go#L106) `v2.6.9` to result anyway.

I think it would be better to save all locations.

knqyf263: OK. When there are two versions for a single package like `debug v3.2.6` and `debug v2.6.9`, how does the location look? I'm concerned that locations will be mixed up.

DmitriyLewen: `debug v3.2.6` and `debug v2.6.9` are 2 different dependencies.
locations for these deps don't mix.
I created tests in Trivy.
[test file](https://github.com/aquasecurity/trivy/blob/952c6dd2f9f78145a194568f0c8d29ec5b2f8630/pkg/fanal/analyzer/language/nodejs/npm/testdata/package-lock.json#L56-L67)
[want](https://github.com/aquasecurity/trivy/blob/952c6dd2f9f78145a194568f0c8d29ec5b2f8630/pkg/fanal/analyzer/language/nodejs/npm/npm_test.go#L88-L114)
Take a look on `ms` dependency.

Tell me if I didn't understand correctly what you meant.

knqyf263: Awesome! Thanks for testing.

knqyf263: I think it is ready for review, no?

DmitriyLewen: oh, yes!

knqyf263 (approved):

PR #138, opened 2022-09-29
feat(swift): add support for cocoapods lock files

DmitriyLewen:
## Description
Add support for swift [cocoapods](https://cocoapods.org/) lock files.


pkg/swift/cocoapods/parse.go
```diff
@@ -0,0 +1,111 @@
...
+	decoder := yaml.NewDecoder(r)
+	if err := decoder.Decode(&lock); err != nil {
+		return nil, nil, xerrors.Errorf("failed to decode cocoapods lock file: %s", err.Error())
+	}
+
+	parsedDeps := map[string]types.Library{} // dependency name => Library
+	directDeps := map[string][]string{}      // dependency name => slice of child dependency names
+	for _, pod := range lock.Pods {
+		switch reflect.ValueOf(pod).Kind() {
+		case reflect.String: // dependency with version number
```
knqyf263, line 40:
```suggestion
		switch pod.(type) {
		case string:
```

pkg/swift/cocoapods/parse.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	var deps []types.Dependency
+	for dep, childDeps := range directDeps {
+		var dependsOn []string
+		// find versions for direct dependencies
+		for _, childDep := range childDeps {
+			// mark this dep as indirect
+			lib := parsedDeps[childDep]
```
knqyf263, line 74:
```
  - AppCenter (4.2.0):
    - AppCenter/Analytics (= 4.2.0)
    - AppCenter/Crashes (= 4.2.0)
  - AppCenter/Analytics (4.2.0):
    - AppCenter/Core
```

In this example, `AppCenter/Analytics` can be direct dependency, no? IIUC, it just says `AppCenter` depends on `AppCenter/Analytics`, but we're not sure if `AppCenter/Analytics` is used directly or not. Please correct me if I'm wrong.

DmitriyLewen: hm... You are right. I didn't thinks about this.
I will change it.

pkg/swift/cocoapods/parse.go
DmitriyLewen: Done

pkg/swift/cocoapods/parse.go
DmitriyLewen: Changed

knqyf263 (approved):

PR #139, opened 2022-10-03
refactor(jar): write debug message when wrong inner jar is found

DmitriyLewen:
## Description
When we found wrong inner jar - we don't need to stop parsing jar.
We need to create debug message and continue parsing.

## Related issues
- aquasecurity/trivy/issues/2952

### DIFF
--- a/pkg/java/jar/parse.go
+++ b/pkg/java/jar/parse.go
@@ -148,7 +148,9 @@ func (p *Parser) parseArtifact(fileName string, size int64, r dio.ReadSeekerAt) 
 		case isArtifact(fileInJar.Name):
 			innerLibs, _, err := p.parseInnerJar(fileInJar) //TODO process inner deps
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/java/jar/parse.go
```diff
@@ -148,7 +148,8 @@ func (p *Parser) parseArtifact(fileName string, size int64, r dio.ReadSeekerAt)
 		case isArtifact(fileInJar.Name):
 			innerLibs, _, err := p.parseInnerJar(fileInJar) //TODO process inner deps
 			if err != nil {
-				return nil, nil, xerrors.Errorf("failed to parse %s: %w", fileInJar.Name, err)
+				log.Logger.Debugf("failed to parse %s: %s", fileInJar.Name, err)
```
knqyf263, line 151: Debug messages should be capitalized.

```suggestion
				log.Logger.Debugf("Failed to parse %s: %s", fileInJar.Name, err)
```

DmitriyLewen: Fixed

knqyf263 (approved):

PR #141, opened 2022-10-13
feat(yarn): add support dependency location for yarn.lock

DmitriyLewen:
## Description
add support dependency location for `yarn.lock` files


pkg/nodejs/yarn/parse_testcase.go
```diff
@@ -8,2312 +8,2312 @@ var (
...
-		{Name: "loose-envify", Version: "1.4.0"},
-		{Name: "object-assign", Version: "4.1.1"},
-		{Name: "promise", Version: "8.0.3"},
-		{Name: "prop-types", Version: "15.7.2"},
-		{Name: "react-is", Version: "16.8.6"},
-		{Name: "react", Version: "16.8.6"},
-		{Name: "redux", Version: "4.0.1"},
-		{Name: "scheduler", Version: "0.13.6"},
-		{Name: "symbol-observable", Version: "1.2.0"},
+		{Name: "asap", Version: "2.0.6", Locations: []types.Location{{StartLine: 5, EndLine: 5}}},
```
knqyf263, line 20: How did you make the expected locations?

DmitriyLewen: `yarn list` command can't get location for dependencies.
I had to `grep` dependency names with versions and paste them into testcase.

knqyf263: If you have the command snippet like `grep | xxx | yyy | ...` extracting locations, please write it down.

DmitriyLewen: i added my commands to get dependencies with location from lock file.
But i worked with [files from testdata](https://github.com/aquasecurity/go-dep-parser/tree/main/pkg/nodejs/yarn/testdata).

During last test for script i found that some dependencies and their versions are different from dependencies and versions when using comments to install deps(i mean [this](https://github.com/aquasecurity/go-dep-parser/blob/5415cc446853b128cafff7230eaf93ddcc1ca45d/pkg/nodejs/yarn/parse_testcase.go#L6-L8)).

I think dependencies can be updated again and our testcase will be not actual.
I can update deps in testcase or remove these instructions and we will only use lock file for tests. 
@knqyf263 wdyt about it?

knqyf263: Yes, I should have used image digests rather than image tags...
I'll merge this PR now and would ask you to update the test data.

knqyf263 (approved):

PR #142, opened 2022-10-17
feat(nuget): add support dependency location for lock files

DmitriyLewen:
## Description
Add support dependency location for nuget lock files


pkg/nuget/lock/parse_test.go
```diff
@@ -52,6 +52,12 @@ func TestParse(t *testing.T) {
 				return ret < 0
 			})
 
+			for _, lib := range got {
+				sort.Slice(lib.Locations, func(i, j int) bool {
```
knqyf263, line 56: What if sorting locations in `Parse`?

DmitriyLewen: hmm.. that's a good idea. 
Sorted locations will be convenient for the user's work.
I will add it.

DmitriyLewen: Done!

PR #143, opened 2022-10-22
feat: add kas parser

nejch:
This is part 1 of adding a few git-based dependencies to trivy - see https://github.com/aquasecurity/trivy/issues/3067. There is more context there, I'm not sure if this is the right place for a git-based parser so I'm happy to make changes, even if this code should live in another repo etc :)

An example real life kas file:
https://github.com/siemens/meta-iot2050/blob/master/kas/iot2050.yml

This PR is largely based on the pnpm parser PR as that had a similar yaml parser, though kas config files are even simpler.
I'll follow up on the trivy side with an analyzer, if this makes sense.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.1 // indirect
 	github.com/liamg/jfather v0.0.7 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/whilp/git-urls v1.0.0 // indirect
 	go.uber.org/atomic v1.7.0 // indirect
 	go.uber.org/multierr v1.6.0 // indirect
 	golang.org/x/text v0.3.3 // indirect
--- a/pkg/git/kas/parse.go
+++ b/pkg/git/kas/parse.go
@@ -1,0 +1,78 @@
+package kas
+
+import (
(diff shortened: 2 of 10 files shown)


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.1 // indirect
 	github.com/liamg/jfather v0.0.7 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/whilp/git-urls v1.0.0 // indirect
```
nejch, line 25: Not sure how you'll feel about adding this little package. But still seemed better than manually parsing, especially if more git-based parsers are added later.

pkg/git/kas/parse.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, repo := range kasFile.Repos {
+		name := getRepoNamefromUri(repo.Url)
+		if name == "" || repo.Url == "" {
+			continue
+		}
+
+		version := repo.RefSpec
+		if version == "" {
+			version = "latest"
+		}
```
nejch, line 57: I add the latest here because without a `refspec`, it will clone HEAD (like implicit `latest` in docker), so just wanted to ensure this still gets picked up for BOMs etc.

PR #144, opened 2022-10-22
feat: add gitlab-ci include parser

nejch:
Very similar to https://github.com/aquasecurity/go-dep-parser/pull/143, this is part 2 of adding a few git-based dependencies to trivy - see https://github.com/aquasecurity/trivy/issues/3067 for more context.

This adds basic support for gitlab-ci includes, mostly for discussion in https://github.com/aquasecurity/trivy/issues/3067. I'll extend/adapt this based on feedback and discussions in the issue, and first see if it makes sense to add it here :)

### DIFF
--- a/pkg/gitlab/include/parse.go
+++ b/pkg/gitlab/include/parse.go
@@ -1,0 +1,64 @@
+package include
+
+import (
+	"golang.org/x/xerrors"
+	"gopkg.in/yaml.v3"
+
+	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
+)
+
+type Include struct {
+	Project string `yaml:"project,omitempty"`
+	Ref     string `yaml:"ref,omitempty"`
+}
+
+type GitlabCiFile struct {
+	Includes []Include `yaml:"include,omitempty"`
+}
+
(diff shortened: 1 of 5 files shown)


nejch:
Sorry for the ping @DmitriyLewen. Are these types of parsers something you'd consider here? If not, I'd maybe look into creating standalone plugins. Thanks a lot!

DmitriyLewen:
Hello @nejch 
Thanks for your work!

I investigated this case.
There is problem with integration this into Trivy.
This parser doesn't match existing categories (i mean languages, os package, etc...).

My opinion - it is better to start from Plugin(as you said). 
If this plugin will have greatest popularity, then we will think about integrating this logic to Trivy.

nejch:
Thanks for your response @DmitriyLewen! 

I understand. I was hoping it would maybe fit into the new `non-packaged` category that was recently introduced for unpackaged executables, see:

https://github.com/aquasecurity/trivy/blob/5bb3a47e03e0156a7949dfd453908798eb8a0b65/pkg/fanal/analyzer/const.go#L77-L80

However, I'll look into packaging this into a plugin as well and maybe open/edit the issue to track popularity if needed.

DmitriyLewen:
I can suggest you(if you have time for this) to create draft PR in Trivy (use these changes in PR) with tests.
Then we can look at this logic, test it and after then decide on the integration of this functionality.

nejch:
That's great, I'll give that a shot :)

PR #148, opened 2022-10-27
feat(yarn): add dependency parsing

VaismanLior:


naortalmor1:
Hi @DmitriyLewen,
Can you take a look on that?

DmitriyLewen (commented):
added some comments

Also questions about testcases(GitHub doesn't want to open this file in browser):
1) I tried use script to get list of Libraries, but got nothing. Can you check it?

2) about script for Dependencies:
can you add line for install `jq` : `apk add jq`?
and add this command + script before every dependency testcase(as for Libraries)

pkg/nodejs/yarn/parse.go
```diff
@@ -2,47 +2,73 @@ package yarn
...
+	"github.com/samber/lo"
 	"golang.org/x/xerrors"
 )
 
 var (
-	yarnLocatorRegexp = regexp.MustCompile(`"?(?P<package>.+?)@(?:(?P<protocol>.+?):)?.+`)
-	yarnVersionRegexp = regexp.MustCompile(`\s+"?version:?"?\s+"?(?P<version>[^"]+)"?`)
+	yarnLocatorRegexp      = regexp.MustCompile(`\\?"?(?P<package>.+?)@(?:(?P<protocol>.+?):)?(?P<version>.+?)\\?"?:?$`)
+	yarnVersionRegexp      = regexp.MustCompile(`\s+"?version:?"?\s+"?(?P<version>[^"]+)"?`)
+	yarnDependenciesRegexp = regexp.MustCompile(`\s+"?dependencies:?"?`)
```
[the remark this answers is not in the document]
VaismanLior: Done

pkg/nodejs/yarn/parse.go
```diff
@@ -2,47 +2,73 @@ package yarn
...
+	packagename, protocol, _, err = parseLocator(locsSplit[0])
+	if err != nil {
+		return "", "", nil, err
+	}
+	locs = lo.FlatMap(locsSplit, func(loc string, _ int) []string {
+		_, _, version, _ := parseLocator(loc)
+		ls := []string{utils.PackageID(packagename, version)}
+		if protocol != "" {
+			ls = append(ls, packagename+"@"+protocol+":"+version)
+		}
```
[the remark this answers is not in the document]
VaismanLior: You are correct, I don't remember why I did this

DmitriyLewen:
@VaismanLior 
Thanks a lot. I am asking about this because i created #147 ising image digest and dependency versions and i need to update it after merging your PR.

VaismanLior:
> @VaismanLior Thanks a lot. I am asking about this because i created #147 ising image digest and dependency versions and i need to update it after merging your PR.

@DmitriyLewen Makes sense, just updated the PR and everything should work now.
I created a Dockerfile for creating the tests cases, that's the only way I could get my script to run

pkg/nodejs/yarn/parse_testcase.go
```diff
@@ -3,19 +3,23 @@ package yarn
...
+	// cat yarn.lock | awk '/^\S+@[~^*]?(>= )?[0-9.]*/,/^$/{if($0=="") {print "--"prev} else { if(substr($0,1,2)!="  ") {print NR":"$0} else {print $0}} prev=NR}; END{print "--"prev}' | awk 'BEGIN {s=""}; {(substr($0,1,2)=="--") ? (s=s$0"\n") : (s=s$0)}; END { print s}' | sed -E 's/@([0-9~><*\^]|npm).*version:? "?/:/' | sed 's/  /:/' | sed 's/"//g'| awk 'match($0, /[[:digit:]]+$/) {print substr($0, RSTART, RLENGTH)":"$0 }' |  awk -F":" '{print "{Name: \""$3"\", Version: \""$4"\", Locations: []types.Location{{StartLine: "$2", EndLine: "$1"}}},"}'
 	yarnNormal = []types.Library{
 		{Name: "asap", Version: "2.0.6", Locations: []types.Location{{StartLine: 5, EndLine: 8}}},
 		{Name: "jquery", Version: "3.4.1", Locations: []types.Location{{StartLine: 10, EndLine: 13}}},
 		{Name: "promise", Version: "8.0.3", Locations: []types.Location{{StartLine: 15, EndLine: 20}}},
 	}
 
-	// yarn list --json --no-progress | jq -r '[ .data.trees[] | { ID: .name, DependsOn: [ .children[] | { name: .name, color: .color } ] } ]'
+	// ... and
+	// yarn --cwd test_deps_generator install
```
DmitriyLewen, line 21: i get error in this command:
```sh
/ # yarn --cwd test_deps_generator install
Error: ENOTDIR: not a directory, open 'test_deps_generator/.yarnrc'
    at Object.openSync (fs.js:498:3)
    at readFileSync (fs.js:394:35)
    at /opt/yarn-v1.22.19/lib/cli.js:97395:58
    at Array.map (<anonymous>)
    at parseRcPaths (/opt/yarn-v1.22.19/lib/cli.js:97393:78)
    at Object.findRc (/opt/yarn-v1.22.19/lib/cli.js:97407:10)
    at getRcConfigForCwd (/opt/yarn-v1.22.19/lib/cli.js:56965:74)
    at buildRcArgs (/opt/yarn-v1.22.19/lib/cli.js:57011:16)
    at getRcArgs (/opt/yarn-v1.22.19/lib/cli.js:57067:16)
    at /opt/yarn-v1.22.19/lib/cli.js:88330:76

```
Can you take a look?

pkg/nodejs/yarn/parse_testcase.go
```diff
@@ -14865,564 +14883,576 @@ var (
...
+	// node test_deps_generator/index.js yarn.lock
 	yarnV2ManyDeps = []types.Dependency{
 		{
-			ID: "promise@8.1.0",
+			ID: "fsevents@2.1.3",
 			DependsOn: []string{
-				"asap@2.0.6",
+				"node-gyp@7.1.0",
 			},
 		},
```
DmitriyLewen, line 14894: i understand correctly = testcases contain dependencies as before but with different order?

pkg/nodejs/yarn/testcase_deps_generator/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM node:14-alpine
+
+COPY index.js /test_deps_generator
+COPY yarn.lock /test_deps_generator
+COPY package.json /test_deps_generator
```
DmitriyLewen, line 5: @knqyf263 wdyt? Is this option okay?

pkg/nodejs/yarn/parse_testcase.go
VaismanLior: Correct 👍🏼

pkg/nodejs/yarn/parse_testcase.go
VaismanLior: Sorry, fixed the dockerfile

pkg/nodejs/yarn/parse_testcase.go
VaismanLior: Correct 👍🏼

pkg/nodejs/yarn/parse_testcase.go
DmitriyLewen: looks like it works good!

DmitriyLewen (approved):

pkg/nodejs/yarn/parse.go
```diff
@@ -60,6 +89,43 @@ func validProtocol(protocol string) (valid bool) {
...
+func parseResults(yarnLibs map[string]Library, dependsOn map[string][]Dependency) (libs []types.Library, deps []types.Dependency) {
+	// find dependencies by locators
+	for libLoc, lib := range yarnLibs {
+		libs = append(libs, types.Library{
+			Name:    lib.Name,
+			Version: lib.Version,
+			Locations: []types.Location{
+				lib.Location,
+			},
+		})
```
DmitriyLewen, line 101: Hello @VaismanLior 
i missed 1 moment: you need to add `ID` here.

DmitriyLewen (changes requested):

pkg/nodejs/yarn/parse.go
VaismanLior: fixed 👍🏼

DmitriyLewen (approved):

pkg/nodejs/yarn/testcase_deps_generator/Dockerfile
DmitriyLewen: We haven't used this way before so I asked you.
Well thank you.

pkg/nodejs/yarn/parse.go
```diff
@@ -69,60 +136,110 @@ func NewParser() types.Parser {
...
 			continue
 		}
+
 		// skip __metadata block
 		if skipPackage = strings.HasPrefix(line, "__metadata"); skipPackage {
 			continue
 		}
+
 		// packagename line start 1 char
 		if line[:1] != " " && line[:1] != "#" {
```
knqyf263, line 212: There are now many states like `isInPackage`, `inDependenciesBlock` and `skipPackage` in a big scope, `Parse`. What if parsing blocks first and then parsing each line instead of having the states?

```
// scanBlocks splits "yarn.lock" by blocks.
func scanBlocks(data []byte, atEOF bool) (advance int, token []byte, err error) {
	if atEOF && len(data) == 0 {
		return 0, nil, nil
	}
	if i := bytes.Index(data, []byte("\n\n")); i >= 0 {
		// We have a full newline-terminated line.
		return i + 2, data[0:i], nil
	} else if i := bytes.Index(data, []byte("\r\n\r\n")); i >= 0 {
		return i + 4, data[0:i], nil
	}

	// If we're at EOF, we have a final, non-terminated line. Return it.
	if atEOF {
		return len(data), data, nil
	}
	// Request more data.
	return 0, nil, nil
}

func (p *Parser) Parse(r dio.ReadSeekerAt) (libs []types.Library, deps []types.Dependency, err error) {
	scanner := bufio.NewScanner(r)
	scanner.Split(scanBlocks) // Split by package blcoks
	for scanner.Scan() {
		block := scanner.Bytes()
		pkg, err := parseBlock(block)
		...
	}
	...
}

func parseBlock(block []byte) {
	scanner := bufio.NewScanner(bytes.NewReader(block))
	var pkg types.Package
	for scanner.Scan() {
		line := scanner.Text()

		// Skip this block
		if strings.HasPrefix(line, "__metadata") {
			return
		}

		// The first line in this package block
		if if line[:1] != " " && line[:1] != "#" {
			parsePackageLocators(line)
			continue
		}

		line = strings.TrimSpace(line)
		switch {
		case strings.HasPrefix(line, "version"):
			pkg.Version, err = parseVersion(line)
		case line == "dependencies:" {
			// start dependencies block
			deps, err := parseDependencies(scanner)
		}
	}
	return pkg, ..., err
}

func parseDependencies(scanner *bufio.Scanner) []types.Dependency {
	var deps []types.Dependency
	for scanner.Scan() {
		line := scanner.Text()
		parseDependency(line)
		...
	}
	return deps
}
```

What do you think of the idea? We call `bufio.NewScanner` twice for splitting blocks and lines, respectively. Then, we don't need to reset those states like `lib = Library{}`, `isInPackage = false` and `inDependenciesBlock = false`. I don't think the above code works, but hopefully, you understand my idea.

We may want to define `ScannerWithLine` to keep the current line number, but it is no big deal. We can manage it somehow.

pkg/nodejs/yarn/parse.go
```diff
@@ -60,69 +92,203 @@ func validProtocol(protocol string) (valid bool) {
...
+
+	// If we're at EOF, we have a final, non-terminated line. Return it.
+	if atEOF {
+		return len(data), data, nil
+	}
+	// Request more data.
+	return 0, nil, nil
+}
+
+func scanProperties(data []byte, atEOF bool) (advance int, token []byte, err error) {
```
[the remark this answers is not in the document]
VaismanLior: I think it's more convenient to parse each property this way, but it definitely makes it harder to count lines so it's a tradeoff. 
If we use `scanLines` we need to keep a flag that tells if we are in the dependencies block(or any other block).
What would you prefer? @knqyf263 
For reference:
 
Screen Shot 2022-12-01 at 10 37 23

pkg/nodejs/yarn/parse.go
```diff
@@ -2,55 +2,87 @@ package yarn
...
-	// TODO : currently yarn can't recognize Dev flag.
-	// That need to parse package.json for Dev flag
-	Dev          bool
-	Dependencies map[string]Dependency
+	Locator string
+	Name    string
 }
 
-func parsePackageLocator(target string) (packagename, protocol string, err error) {
+var lineNumber int
```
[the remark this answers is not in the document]
VaismanLior: Removed 👍🏼

pkg/nodejs/yarn/parse.go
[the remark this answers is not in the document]
knqyf263: The scope is small enough now, so I'm ok with using a flag. But I think you can also pass the line scanner to `parseDependencies`.

```
type LineScanner struct {
	*bufio.Scanner
	lineNum int
}

func NewLineScanner(r io.Reader) *LineScanner {
	return &LineScanner{Scanner: bufio.NewScanner(r)}
}

func (s *LineScanner) Scan() bool {
	s.lineNum++
	return s.Scanner.Scan()
}

func (s *LineScanner) LineNum() int {
	return s.lineNum
}

func parseBlock(block []byte) (Library, []Dependency, int, error) {
	var (
		lib  Library
		deps []Dependency
		err  error
	)
	scanner := NewLineScanner(bytes.NewReader(block))
	for scanner.Scan() {
		line := scanner.Text()

		if len(line) == 0 || line[0] == '#' {
			continue
		}

		...(snip)...

		switch {
		case strings.HasPrefix(line, "dependencies:"):
			// start dependencies block
			if deps, err = parseDependencies(scanner); err != nil {
				return Library{}, nil, 0, err
			}
			continue
		}
	}

	return lib, deps, scanner.LineNum(), nil
}

func parseDependencies(scanner *LineScanner) (deps []Dependency, err error) {
	for scanner.Scan() {
		line := scanner.Text()
		if !strings.HasPrefix(line, strings.Repeat(" ", 4)) { // Lior might have a better idea
			// Get in another property
			break
		}

		if dep, err := parseDependency(line); err != nil {
			return deps, err
		} else {
			deps = append(deps, dep)
		}
	}

	return
}
```

knqyf263 (commented):
I left some comments.

knqyf263 (commented):
I'd suggest simplifying `parseResults`.

knqyf263:
https://github.com/aquasecurity/trivy/issues/3270

pkg/nodejs/yarn/parse.go
```diff
@@ -60,69 +114,189 @@ func validProtocol(protocol string) (valid bool) {
...
+		return len(data), data, nil
+	}
+	// Request more data.
+	return 0, nil, nil
+}
+
+func parseBlock(block []byte, lineNum int) (lib Library, deps []Dependency, newLine int, err error) {
+	var (
+		emptyLines int // lib can start with empty lines first
+		skipBlock  bool
```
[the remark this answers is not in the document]
VaismanLior: Sometimes we need to skip a block but we can't break the loop because the scanner also counts lines number of the block

pkg/nodejs/yarn/parse.go
```diff
@@ -60,69 +114,189 @@ func validProtocol(protocol string) (valid bool) {
...
+	unique := map[string]struct{}{}
+	dependsOn := map[string][]Dependency{}
+	yarnLibs := map[string]Library{}
+	for scanner.Scan() {
+		block := scanner.Bytes()
+		lib, deps, newLine, err := parseBlock(block, lineNumber)
+		lineNumber = newLine + 2
+		if err == nil && lib.Name != "" {
+			symbol := utils.PackageID(lib.Name, lib.Version)
+			if _, ok := unique[symbol]; ok {
```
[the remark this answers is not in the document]
VaismanLior: removed

pkg/nodejs/yarn/parse.go
```diff
@@ -60,69 +114,189 @@ func validProtocol(protocol string) (valid bool) {
...
+
+func parseDependencies(scanner *LineScanner) (deps []Dependency) {
+	for scanner.Scan() {
+		line := scanner.Text()
+		if dep, err := parseDependency(line); err != nil {
+			// finished dependencies block
+			return deps
+		} else {
+			deps = append(deps, dep)
+		}
```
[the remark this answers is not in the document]
VaismanLior: The only error it can get is an expected custom error, so no real need for this.
I can change it to a boolean or use an empty string to make it clearer.

VaismanLior:
Thanks for your suggestions! As always, it looks better now.

knqyf263 (approved):
Thanks for your patience. We iterated reviews many times. It looks very good now.

knqyf263:
Please don't forget to bump go-dep-parser in Trivy and update the docs.

- go.mod
  - https://github.com/aquasecurity/trivy/blob/c110c4e028d50a2253c1d18432b075c5f83a0d34/go.mod#L12
- Docs
  - Dependency location
    - yarn.lock is now supported
    - https://github.com/aquasecurity/trivy/blob/main/docs/docs/vulnerability/detection/language.md
  - Dependency Origin Tree
    - yarn.lock is now supported
    - https://github.com/aquasecurity/trivy/blob/main/docs/docs/vulnerability/examples/report.md#show-origins-of-vulnerable-dependencies

You can build the doc by `make mkdocs-serve` and check it on your browser.

PR #152, opened 2022-11-07
feat(nuget): add dependency parsing

guyb1:


pkg/nuget/lock/parse.go
```diff
@@ -49,7 +51,33 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 				Version: packageContent.Resolved,
 			}
 			libs = append(libs, lib)
+
+			depId := utils.PackageID(packageName, packageContent.Resolved)
+			dependsOn := make([]string, 0)
```
DmitriyLewen, line 56:
```suggestion
			var dependsOn []string
```
I think we can use this variable declaration

pkg/nuget/lock/parse.go
```diff
@@ -49,7 +51,33 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+			}
+
+			if depsMap[depId] != nil {
+				dependsOn = append(dependsOn, depsMap[depId]...)
+				if dependsOn = utils.UniqueStrings(dependsOn); dependsOn == nil {
+					dependsOn = make([]string, 0)
+				}
+			}
+
+			depsMap[depId] = dependsOn
```
DmitriyLewen, line 69: We only need to keep dependencies with child dependencies([e.g. for npm](https://github.com/aquasecurity/go-dep-parser/blob/60502daef4babbc9eee9c57c524625f6f50891f4/pkg/nodejs/npm/parse_testcase.go#L16-L21)).

pkg/nuget/lock/parse.go
```diff
@@ -49,7 +51,33 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+			for depName := range packageContent.Dependencies {
+				dependsOn = append(dependsOn, utils.PackageID(depName, targetContent[depName].Resolved))
+			}
+
+			if depsMap[depId] != nil {
+				dependsOn = append(dependsOn, depsMap[depId]...)
+				if dependsOn = utils.UniqueStrings(dependsOn); dependsOn == nil {
+					dependsOn = make([]string, 0)
+				}
+			}
```
DmitriyLewen, line 67:
```suggestion
			if savedDependsOn, ok := depsMap[depId]; ok {
				dependsOn = utils.UniqueStrings(append(dependsOn, savedDependsOn...))
			}
```
we don't need to check for emptiness (see previous comment)

pkg/nuget/lock/parse.go
guyb1: Actually, that's what I do here :)

DmitriyLewen: you save all dependencies here.
e.g.:
```
...
             "pkgA": {
                "type": "Direct",
                "requested": "[1.0.0, )",
                "resolved": "1.0.0",
                "contentHash": ""
            },
            "pkgB": {
                "type": "Direct",
                "requested": "[2.0.0, )",
                "resolved": "2.0.0",
                "contentHash": "",
                "dependencies": {
                    "pkgC": "3.0.0",
                }
            },
           "pkgC": {
                "type": "Direct",
                "requested": "[3.0.0, )",
                "resolved": "3.0.0",
                "contentHash": ""
            },
...
```
For this lock file you must skip `pkgA` and `pkgC` dependencies, because they have no child deps.
DependOn result:
```
Dependency{
		{ID: "pkgB@2.0.0", DependsOn: []string{"pkgC@3.0.0"}},,
	}
```
I think something like this should be added:
```
if len(dependsOn) > 0 {
	depsMap[depId] = dependsOn
}
```

DmitriyLewen (approved):
LGTM, Thanks for your work!

pkg/nuget/lock/parse.go
```diff
@@ -60,9 +62,34 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 				},
 			}
 			libs = append(libs, lib)
+
+			depId := utils.PackageID(packageName, packageContent.Resolved)
```
knqyf263, line 66: You also need to pass the ID to `types.Library` here.
https://github.com/aquasecurity/go-dep-parser/blob/5545d9feecb875ce2c9f0ea56d7a9a726ed06438/pkg/nuget/lock/parse.go#L54-L63

knqyf263 (approved):

DmitriyLewen:
Hello @guyb1 
Can you also create PR for Trivy?
You need to update `go-dep-parser` version(please use 0f27198c8fbaef01ab7c1f2049c96dd5c9a4b2f6 commit for this) and add/correct tests.
P.S. [Some information](https://aquasecurity.github.io/trivy/v0.34/community/contribute/pr/) about Trivy PR rules.

Thanks in advance!

DmitriyLewen:
UPD:
I created https://github.com/aquasecurity/trivy/pull/3222. Tell me - if i missed something.

PR #154, opened 2022-12-01
fix(go): skip dep without Path for go-binaries

DmitriyLewen:
## Description
Binaries with old go version may incorrectly add module in Deps.
e.g.:
Screenshot from 2022-12-01 14-00-51
Skip dependencies without `Path`.

## Related Issues
- aquasecurity/trivy/issues/3237


knqyf263 (approved):

knqyf263:
Nice catch!

PR #157, opened 2022-12-19
fix(pom): use props from child in parent

DmitriyLewen:
## Description
We currently do not use properties from the base pom.xml in its parent element.
e.g. In this example, we need to use version `2.14.0-rc1`.
image

## Related issues
- aquasecurity/trivy/issues/2145


knqyf263:
I thought properties are propagated from parents to children. Does the diagram mean we have to use child's properties in the parent?

DmitriyLewen:
right. We may want to change some properties of parent in base pom.xml

knqyf263:
Do you have any document? I thought the opposite. I want to clarify what the right behavior is.

DmitriyLewen:
> Do you have any document? 

unfortunately maven doesn't contains detailed information about all logic for pom files.
I played with `mvn dependency tree` after users found differences with trivy and mvn logic.

___

> I thought the opposite

we need to separate 2 cases:
1) base pom.xml has parent struct(as in PR description). In this case:
```
base pom.xml - child
parent - parent
```
**child > parent**
We need to use properties from base pom.xml first.
e.g. Users overwrite version of jackson-bom to fix vulnerabilities.(https://github.com/aquasecurity/trivy/issues/2145#issuecomment-1342964210)

2) base pom.xml has child(dependency struct). In this case:
```
base pom.xml - parent
child - child
```
**parent > child**
This is our current logic.
We need to be able to change properties in the child dependency.

knqyf263:
>Now we may add values to the parent POM, which will be inherited by its children. Most elements from the parent POM are inherited by its children, including:

https://maven.apache.org/pom.html#inheritance

The document seems to say the opposite. The children inherit the parent's properties. Am I missing something?

DmitriyLewen:
I think they mean case, when only parent has some property
e.g.:


But when both parent and base pom have same property - base pom has higher priority:

knqyf263:
Your diagram says `parent-child pom`, but it is not child, right? It seems to be a dependency. Does Maven call a dependency child? I want to explicitly distinguish children and dependencies.

DmitriyLewen:
I understand that i am confused in these terms.
I think it would be better to show an example with real maven packages:
___
i created pom file with `jackson-databind` dependency - https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-without-overwrite-version-pom-xml
version of `jackson-databind` is taken from [parent field](https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-without-overwrite-version-pom-xml-L5-L9)(`jackson-bom`):
`jackson-bom` uses `${jackson.version.databind}` property as version(https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-copy-of-jackson-bom-2-13-4-pom-L90-L94).
tthere is following property chain:
[${jackson.version.databind}](https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-copy-of-jackson-bom-2-13-4-pom-L62) <- [jackson-version](https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-copy-of-jackson-bom-2-13-4-pom-L52) <- `v2.13.4`
for this pom file `Maven` uses `2.13.4` version:
```
➜ mvn dependency:tree
...
[INFO] --------------------------< com.example:test >--------------------------
[INFO] Building test 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:3.1.2:tree (default-cli) @ test ---
[INFO] com.example:test:jar:0.0.1-SNAPSHOT
[INFO] \- com.fasterxml.jackson.core:jackson-databind:jar:2.13.4:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.4:compile
[INFO]    \- com.fasterxml.jackson.core:jackson-core:jar:2.13.4:compile
...
```

___
After that i add `{jackson.version.databind}` property in base pom.xml - https://gist.github.com/DmitriyLewen/35aa0573d8f97deb0ddee5f20a728dfe#file-with-overwrite-version-pom-xml-L15
in this case, we need to take this property from base pom.xml and use it in parent (j`ackson-bom`).
Then `Maven` uses `2.14.0-rc1` for `jackson-databind`:
```
➜ mvn dependency:tree
...
[INFO] --------------------------< com.example:test >--------------------------
[INFO] Building test 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:3.1.2:tree (default-cli) @ test ---
[INFO] com.example:test:jar:0.0.1-SNAPSHOT
[INFO] \- com.fasterxml.jackson.core:jackson-databind:jar:2.14.0-rc1:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.4:compile
[INFO]    \- com.fasterxml.jackson.core:jackson-core:jar:2.13.4:compile
...
```

knqyf263:
After a while, I figured out the reason. You tried to propagate child properties to parents, but it was not the correct approach if I understand correctly. Our mistake is evaluating properties in parents. We need to keep original values in dependencyManagement and evaluate them after merging properties. It makes sense child properties take precedence over parent properties.

In your example, we keep `version` without evaluating.

```
<version>${jackson.version.databind}</version>
```

Then, merge properties `jackson.version.databind`:
- parent: 2.13.4
- child: 2.14.0-rc1

=> 2.14.0-rc1 is preferred

At last, we evaluate `${jackson.version.databind}`. The property is now `2.14.0-rc1`.

All we have to do is remove this line.
https://github.com/aquasecurity/go-dep-parser/pull/157/files#diff-00cf648cd617725ff0a56f14bda5522654f75e04db3f460fc747050db59a42a3L273-L275

knqyf263:
I've started changing it to accept `fs.FS` for `data` according to Chen's suggestion.

DmitriyLewen:
I tested your changes and looks like this is doesn't work for case when props used not for dependency in child pom.xml(e.g. bom files).
example that you right understand what i mean:

parent(`spring-boot-dependencies-2.7.4.pom` - https://gist.github.com/DmitriyLewen/65ffbc183aac0da996d7f3021c79e3ae#file-with-overwrite-bom-version-L5-L9) has no property with `jackson-databind` version. But it [has jackson-bom](https://gist.github.com/DmitriyLewen/65ffbc183aac0da996d7f3021c79e3ae#file-spring-boot-dependencies-2-7-4-pom-L19-L25). We should get `jackson-databind` version from this bom file.(`jackson-bom` use own version for `jackson-databind` - https://gist.github.com/DmitriyLewen/65ffbc183aac0da996d7f3021c79e3ae#file-jackson-bom.)
In your changes Parent doesn't have information about new bom version and still use `2.13.4` version(https://gist.github.com/DmitriyLewen/65ffbc183aac0da996d7f3021c79e3ae#file-spring-boot-dependencies-2-7-4-pom-L13):
```
2023-02-21T09:37:09.592+0600    DEBUG   Resolving com.fasterxml.jackson:jackson-bom:2.13.4...

```

If we overwrite bom version() - maven will use new `jackson-databind` version:
```sh
➜  with mvn dependency:tree
[INFO] Scanning for projects...
...
[INFO] 
[INFO] --------------------------< com.example:test >--------------------------
[INFO] Building test 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:3.3.0:tree (default-cli) @ test ---
[INFO] com.example:test:jar:0.0.1-SNAPSHOT
[INFO] \- com.fasterxml.jackson.core:jackson-databind:jar:2.14.0-rc1:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.14.0-rc1:compile
[INFO]    \- com.fasterxml.jackson.core:jackson-core:jar:2.14.0-rc1:compile
[INFO] ------------------------------------------------------------------------
...
```

without overwrite():
```sh
➜  without mvn dependency:tree
[INFO] Scanning for projects...
...
[INFO] 
[INFO] --------------------------< com.example:test >--------------------------
[INFO] Building test 0.0.1-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- maven-dependency-plugin:3.3.0:tree (default-cli) @ test ---
[INFO] com.example:test:jar:0.0.1-SNAPSHOT
[INFO] \- com.fasterxml.jackson.core:jackson-databind:jar:2.13.4:compile
[INFO]    +- com.fasterxml.jackson.core:jackson-annotations:jar:2.13.4:compile
[INFO]    \- com.fasterxml.jackson.core:jackson-core:jar:2.13.4:compile
[INFO] ------------------------------------------------------------------------
...
```

___

Also there are problems with resolving some packages - missing version of packages:
```
2023-02-21T09:37:09.618+0600    DEBUG   Resolving org.springframework.session:spring-session-bom:2021.2.0...
2023-02-21T09:37:09.642+0600    DEBUG   Resolving com.fasterxml.jackson.core:jackson-databind:...
2023-02-21T09:37:10.012+0600    DEBUG   com.fasterxml.jackson.core:jackson-databind: was not found in local/remote repositories
```

knqyf263:
It looks super weird. Are properties in dependencyManagement inherited? jackson-bom's properties should not affect, no?

knqyf263:
I'm suppect that spring-boot-dependencies-2.7.4.pom has jacson-databind in dependencyManagement.

DmitriyLewen:
> It looks super weird

I thought so too.
But next time i understood next logic:
You create `bom` files(`jackson-bom`) with all properties and deps required to your project.
and use this `bom` in parent pom(`spring-boot-dependencies`).

If you need overwrite 1 properties - you can add new value in `properties` field in base `pom.xml`.
But in case when you need to overwrite some props => you can update your `bom` file.
There may be reason not to update parent, and you should be able to overwrite `bom` file version without changing parent.

> I'm suppect that spring-boot-dependencies-2.7.4.pom has jacson-databind in dependencyManagement.

there is next chain:
`spring-boot-dependencies2.7.4.pom` -> `jackson-bom.${jackson-bom.version}.pom`(with `import` scope) -> `jackson-databind.${jackson.version.databind}.pom`

By default spring-boot-dependencies2.7.4.pom has ${jackson-bom.version} == 2.13.4.
If we overwite ${jackson-bom.version}, then we get new version of  `jackson-databind`

knqyf263:
Thanks for adding the test. This is my understanding.

image

knqyf263 (approved):

PR #160, opened 2022-12-28
fix(yarn): yarn parser error handling

VaismanLior:
Found an issue when failed to parse a dependency block:
[This](https://github.com/aquasecurity/go-dep-parser/blob/main/pkg/nodejs/yarn/parse.go#L214) line causes the scanner to return a result altho it should've thrown an error.

### DIFF
--- a/pkg/nodejs/yarn/parse.go
+++ b/pkg/nodejs/yarn/parse.go
@@ -205,7 +205,7 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine 
 			if patterns == nil || !validProtocol(protocol) {
 				skipBlock = true
 				if !ignoreProtocol(protocol) {
-					err = xerrors.Errorf("failed to parse package patterns")
+					return Library{}, nil, -1,  xerrors.Errorf("failed to parse package patterns")
 				}
 				continue
 			} else {
--- a/pkg/nodejs/yarn/parse.go
+++ b/pkg/nodejs/yarn/parse.go
@@ -105,10 +105,18 @@ func getDependency(target string) (name, version string, err error) {
 	return capture[1], capture[2], nil
 }
 
-func validProtocol(protocol string) (valid bool) {
+func validProtocol(protocol string) bool {
 	switch protocol {
 	// only scan npm packages
 	case "npm", "":
+		return true
+	}
+	return false
+}
+
+func ignoreProtocol(protocol string) bool {
+	switch protocol {
+	case "workspace", "patch":
 		return true
 	}
 	return false
@@ -196,7 +204,9 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine 
 		if name, protocol, patterns, patternErr := parsePackagePatterns(line); patternErr == nil {
 			if patterns == nil || !validProtocol(protocol) {
 				skipBlock = true
-				err = xerrors.Errorf("failed to parse package patterns")
+				if !ignoreProtocol(protocol) {
+					err = xerrors.Errorf("failed to parse package patterns")
+				}
 				continue
 			} else {
 				lib.Patterns = patterns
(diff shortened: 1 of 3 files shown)


pkg/nodejs/yarn/parse.go
```diff
@@ -196,7 +204,9 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine
 		if name, protocol, patterns, patternErr := parsePackagePatterns(line); patternErr == nil {
 			if patterns == nil || !validProtocol(protocol) {
 				skipBlock = true
-				err = xerrors.Errorf("failed to parse package patterns")
+				if !ignoreProtocol(protocol) {
+					err = xerrors.Errorf("failed to parse package patterns")
```
knqyf263, line 208: If we don't want to fail on this error, we can just show warnings.

```suggestion
					log.Logger.Warnf("Unsupported protocol: %s", protocol)
```

If we want to fail on this error, we should return an error here.

```suggestion
					return Library{}, nil, -1,  xerrors.Errorf("failed to parse package patterns")
```

VaismanLior: I lean towards failing to keep it compatible with what we had before the refactor

knqyf263 (approved):

jerbob92:
This change seems to break the scanning when the `yarn.lock` contains the file protocol (packages directly installed from a local tar.gz file), perhaps we should add `file` to `ignoreProtocol`?

knqyf263:
@jerbob92 Oh, it needs to be fixed soon. Thanks for letting us know. @VaismanLior What do you think?

>Also not quite sure if I agree with completely quitting the scanning when it can't process 1 dependency.

Agree with you. Warning sounds better.

VaismanLior:
Does a file dependency have a version that we can extract from it? Just for visibility, to know that it is part of the project.
Or maybe we can put the file path as the version?
What would you expect in your use case? @jerbob92 

I guess ignoring it is an option too... As trivy can't really audit vulnerabilities for it.
The reason I went with failing is for the visibility of the dependency instead of skipping it completely. But I'm good with a warning instead if we find a better idea.

jerbob92:
> Does a file dependency have a version that we can extract from it? Just for visibility, to know that it is part of the project.
Yes it has a version that can be extracted

> What would you expect in your use case? @jerbob92
I'm not really sure, I just tried adding `file` to `validProtocol()` and that also works, since it does have a pattern (the pattern is the file path). Not sure what Trivy does with it down the line.

> I guess ignoring it is an option too... As trivy can't really audit vulnerabilities for it.
Ignoring is the best way I guess, even if the package name matches a package on npm, and the version exists, and it has vulnerabilities, it doesn't really mean anything because the user could have done anything to the file the package points to.

> The reason I went with failing is for the visibility of the dependency instead of skipping it completely. But I'm good with a warning instead if we don't have a better idea.
I think failing would be OK if it fails on actual issues in the lockfile, and that it cascades the failure completely to the scan, making the command exit with an exit code, but it doesn't right now. So probably a warning would be better for now.

knqyf263:
I was thinking about SBOM. We may want to add those dependencies to SBOM even though they cannot be used for vulnerabilities.

>Ignoring is the best way I guess, even if the package name matches a package on npm, and the version exists, and it has vulnerabilities, it doesn't really mean anything because the user could have done anything to the file the package points to.

It may lead to false detection like @jerbob92 pointed out. We can add a new field for protocol and skip dependencies that are not installed from npm registries, but we can think of it in the future. We should fix this error for now.

@jerbob92 Do you mind creating PR? You already have a lock file failing the parser. Or if you share the lock file with us, we can fix that.

jerbob92:
Here's a `yarn.lock` that will trigger it: 

```
"local-package@file:vendor/local-package.tar.gz":
  version "0.0.1"
  resolved "file:vendor/local-package.tar.gz#f0b5161c241e8c94acc6241c49e3ab6806e6d4d7"
```

Might make a PR later today if I can find the time.

knqyf263:
Thanks for your help. Please let us know if you don't have time. We'll take care of it.

PR #161, opened 2023-01-04
Ignore dependencies with file protocol and add a test case for it

jerbob92:
Fixes a regression that was introduced in #160.

This change will ignore dependencies that reference a local file path. While these may have a version, it does not mean that the package can be resolved on NPM or that that version that's in the lock file is the actual version of the package, meaning that this could lead to false detections. 

We may want to add these dependencies for SBOM later on, we can do this by moving `file` from `ignoreProtocol()` to `validProtocol()` and mark the package as being a file reference so that Trivy doesn't try to match it to NPM packages.


jerbob92:
Here you go @knqyf263

knqyf263 (approved):
Thanks for your awesome work!

knqyf263:
Do you have time to create a PR just bumping go-dep-parser in Trivy? If no, we'll do.

PR #163, opened 2023-01-15
fix(yarn): ignore more protocols

VaismanLior:
Fixed a trivy issue https://github.com/aquasecurity/trivy/issues/3428
Added more protocols to the ignore list and more details in the parse error.
For reference: https://yarnpkg.com/features/protocols


knqyf263 (approved):

knqyf263:
Can you also bump go-dep-parser in trivy?

VaismanLior:
Will do

PR #169, opened 2023-02-14
feat(pipenv): add dependency locations for Pipfile.lock

DmitriyLewen:
## Description
Add dependency locations for `Pipfile.lock`

### DIFF
--- a/pkg/python/pipenv/parse.go
+++ b/pkg/python/pipenv/parse.go
@@ -1,7 +1,8 @@
 package pipenv
 
 import (
-	"encoding/json"
+	"github.com/liamg/jfather"
+	"io"
 	"strings"
 
 	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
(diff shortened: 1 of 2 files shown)


pkg/python/pipenv/parse.go
```diff
@@ -10,11 +11,13 @@ import (
 )
 
 type lockFile struct {
-	Default map[string]dependency
-	Develop map[string]dependency
+	Default map[string]dependency `json:"default"`
+	//Develop map[string]dependency `json:"develop"`
```
knqyf263, line 15: Do we want to remove it?

DmitriyLewen: We don't use developer dependencies, so i think we can remove it.

DmitriyLewen: Done

knqyf263 (approved):

PR #172, opened 2023-02-21
feat(ruby): dependency tree support for Gemfile.lock

AndreyLevchenko:
Add dependency tree support for Gemfile.lock


DmitriyLewen (approved):

DmitriyLewen:
@AndreyLevchenko `Gemfile.lock` file always has only 1 versions of library, right?

AndreyLevchenko:
> @AndreyLevchenko `Gemfile.lock` file always has only 1 versions of library, right?

From what I can understand `bundler` either resolve dependency to single version or throw "Bundler could not find compatible versions for gem ..." error

knqyf263:
We need to determine direct dependencies. I've updated.
v

PR #176, opened 2023-02-24
feat(gomod): add package id

knqyf263:


pkg/golang/mod/parse.go
```diff
@@ -138,3 +153,7 @@ func lessThan117(ver string) bool {
 
 	return major <= 1 && minor < 17
 }
+
+func pkgID(name, version string) string {
+	return fmt.Sprintf("%s@v%s", name, version)
+}
```
DmitriyLewen, line 159: i am not sure that understand - why do we need `v`version prefix? 
I mean why don't we use this function:
https://github.com/aquasecurity/go-dep-parser/blob/f6e7eca870438a14ea2d477660227cc5298c3a89/pkg/utils/utils.go#L55

pkg/golang/sum/parse_test.go
```diff
@@ -39,10 +39,15 @@ func TestParse(t *testing.T) {
...
 			f, err := os.Open(v.file)
 			require.NoError(t, err)
+			defer f.Close()
 
 			got, _, err := NewParser().Parse(f)
 			require.NoError(t, err)
 
+			for i := range got {
+				got[i].ID = "" // Not compare IDs
+			}
```
DmitriyLewen, line 49: we can use `go list -m all | awk 'NR>1 {sub(/^v/, "", $2); printf("{ID: \""$1"@v"$2"\", Name: \""$1"\", Version: \""$2"\"},\n")}'
` to create ID.
Or is there another reason to skip  `IDs` check?

pkg/golang/mod/parse.go
knqyf263: Added a comment
https://github.com/aquasecurity/go-dep-parser/pull/176/commits/01065334947fdd71efb759ec46cc28eb7cd1f3f1

Go has their ID format like `<module_name>@v<module_version>`. It is better to comply with that so we can easily identify the module path like `$GOPATH/pkg/mod/github.com/aquasecurity/go-dep-parser@v0.0.0-20230130190635-5e31092b0621`. You can see the `v` prefix.

pkg/golang/sum/parse_test.go
knqyf263: ID is like a pseudo field generated by us. In this test, I just want to compare the fields shown by the Go official tool like `go list`. I've added another test.
https://github.com/aquasecurity/go-dep-parser/pull/176/commits/132ab47235b202895d1e3e220032ac62e0c4c8c5

DmitriyLewen (approved):

Issue #179, opened 2023-02-28
add dependency tree and line numbers to rust cargo lock files


AndreyLevchenko:

knqyf263:
https://github.com/aquasecurity/go-dep-parser/pull/180

PR #180, opened 2023-02-28
feat(rust): dependency tree and line numbers support for cargo lock file (#179)

AndreyLevchenko:


pkg/rust/cargo/parse_test.go
```diff
@@ -37,27 +72,40 @@ func TestParse(t *testing.T) {
...
 }
+
+func sortLibs(libs []types.Library) {
+	sort.Slice(libs, func(i, j int) bool {
+		ret := strings.Compare(libs[i].Name, libs[j].Name)
+		if ret == 0 {
+			return libs[i].Version < libs[j].Version
+		}
+		return ret < 0
+	})
```
DmitriyLewen, line 104:
```suggestion
	sort.Slice(libs, func(i, j int) bool {
		return strings.Compare(libs[i].ID, libs[j].ID) < 0
	})
```
looks like we can just use `ID`.

pkg/rust/cargo/parse.go
```diff
@@ -9,13 +12,14 @@ import (
...
+}
 type Lockfile struct {
-	Packages []struct {
-		Name         string   `toml:"name"`
-		Version      string   `toml:"version"`
-		Source       string   `toml:"source,omitempty"`
-		Dependencies []string `toml:"dependencies,omitempty"`
-	} `toml:"package"`
+	Packages []cargoPkg `toml:"package"`
 	Metadata interface{}
```
DmitriyLewen, line 23: If we still use `toml` decoder:

we don't use tag for `Metadata`, but toml decoder still parses this field:
image
Maybe it make sense to remove `Metadata`?

pkg/rust/cargo/parse.go
```diff
@@ -27,18 +31,57 @@ func NewParser() types.Parser {
 
 func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
 	var lockfile Lockfile
-	if _, err := toml.DecodeReader(r, &lockfile); err != nil {
+	decoder := toml.NewDecoder(r)
+	if _, err := decoder.Decode(&lockfile); err != nil {
 		return nil, nil, xerrors.Errorf("decode error: %w", err)
 	}
```
DmitriyLewen, line 37: we will parse lock file again(i mean `naivePkgParser`).
Perhaps it make sense to remove this decoding and take all dependency information from `naivePkgParser`?

pkg/rust/cargo/parse.go
```diff
@@ -27,18 +31,57 @@ func NewParser() types.Parser {
...
+		if dep != nil {
+			deps = append(deps, *dep)
+		}
+	}
+	return libs, deps, nil
+}
+func parseDependencies(pkgId string, pkg cargoPkg) *types.Dependency {
+	dependOn := []string{}
+	for _, pkgDep := range pkg.Dependencies {
+		fields := strings.Fields(pkgDep)
```
DmitriyLewen, line 71: Can you add comment about this field schema?
if i understand correctly it is: `<pkg_name> <pkg_version> (registry+<link>)` , right?

pkg/rust/cargo/naive_pkg_parser.go
```diff
@@ -0,0 +1,79 @@
...
+			if currentPkg.name != "" {
+				pkgId := utils.PackageID(currentPkg.name, currentPkg.version)
+				currentPkg.setEndPositionIfEmpty(lineNum - 1)
+				idx[pkgId] = currentPkg.position
+			}
+			currentPkg = minPkg{}
+			currentPkg.position.start = lineNum
+
+		} else if matched := pkgNameRegexp.MatchString(line); matched {
+			currentPkg.name = propertyValue(line)
```
DmitriyLewen, line 56: There are times when a simple regular expression leaks memory.
What if we remove all `regexp` and use `strings` package?
Something like this:
```suggestion
		} else if strings.HasPrefix(strings.TrimSpace(line), "name") {
			currentPkg.name = propertyValue(line)
```

DmitriyLewen (approved):
Good work! 
Thank you.

pkg/rust/cargo/testdata/cargo_invalid.lock
```diff
@@ -0,0 +1,34 @@
...
+dependencies = [
+ "libc 0.2.54 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "typemap"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unsafe-any 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
knqyf263, line 20: `unsafe-any` is not on the package list. Is it possible?

knqyf263: OK, this is an invalid case, but how about `pkg/rust/cargo/testdata/cargo_normal.lock` and `pkg/rust/cargo/testdata/cargo_mixed.lock`? `unsafe-any` is in dependencies, but I didn't find it in package list. I want to use a valid cargo.lock for valid cases.

knqyf263 (approved):

PR #182, opened 2023-03-01
Support for optional dependencies in pyproject

ilpianista:
Should help fixing aquasecurity/trivy#3720

### DIFF
--- a/pkg/python/pyproject/pyproject.go
+++ b/pkg/python/pyproject/pyproject.go
@@ -1,9 +1,10 @@
 package pyproject
 
 import (
+	"io"
+
 	"github.com/BurntSushi/toml"
 	"golang.org/x/xerrors"
-	"io"
 )
 
 type PyProject struct {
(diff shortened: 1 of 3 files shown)


ilpianista:
> Thanks for your contribution! Could you keep the existing test as is and add a new test for this case? Then, we can make sure it works with both string and object.

Well, I didn't remove the existing dependencies (which are in the form `<name>:<string>`), thus this should guarantee that it works with both types. Or am I missing anything?

DmitriyLewen:
Hello @ilpianista 
Thanks for your work!

i found 1 more case - dependency is array(https://github.com/python-poetry/poetry/blob/3e32a4f83e89b2b2bede68e0d375a70931b04a8e/pyproject.toml#L80-L84)
can you also add this case in test?

ilpianista:
@DmitriyLewen done.

knqyf263:
>Well, I didn't remove the existing dependencies (which are in the form <name>:<string>), thus this should guarantee that it works with both types. Or am I missing anything?

It is ok, then. Thanks!

knqyf263 (approved):

PR #184, opened 2023-03-02
merge

MorAlon1:


VaismanLior:
Please update the title to be informative, and add tests for the change

pkg/nodejs/yarn/parse.go
```diff
@@ -297,5 +306,33 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	// Replace dependency patterns with library IDs
 	// e.g. ajv@^6.5.5 => ajv@6.10.0
 	deps := parseResults(patternIDs, dependsOn)
+
+	for i, lib := range libs {
```
VaismanLior, line 310: I think we don't need this logic.
The `parseResults` func already iterates all the dependencies of the yarn file.
We can create a unique set of all the indirect dependencies and return it from `parseResults`.
Then simply check if the `lib.ID` is in the set

knqyf263:
No activity and no description. If it is still relevant, we can open a new PR with more information.

PR #185, opened 2023-03-03
feat(packagejson): parse dependencies/optionalDependencies

DmitriyLewen:
## Description
We need dependencies from `package.json` files to update dependencies from `yarn.lock` files.

### DIFF
--- a/pkg/nodejs/packagejson/parse.go
+++ b/pkg/nodejs/packagejson/parse.go
@@ -10,14 +10,12 @@ import (
 )
 
 type packageJSON struct {
-	Name    string      `json:"name"`
-	Version string      `json:"version"`
-	License interface{} `json:"license"`
-}
-
-type dependencies struct {
+	Name         string            `json:"name"`
+	Version      string            `json:"version"`
+	License      interface{}       `json:"license"`
 	Dependencies map[string]string `json:"dependencies"`
 }
+
 type Parser struct{}
 
 func NewParser() types.Parser {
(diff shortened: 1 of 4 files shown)


knqyf263 (commented):
I was thinking of changing the interface as below. It allows having the information of the project itself. We need this change for Go modules and Pipenv as well.

```
type Parser interface {
	// Parse parses the dependency file
	Parse(r dio.ReadSeekerAt) (Library, []Library, []Dependency, error)
}
```

Is it too big a change?

DmitriyLewen:
It's not hard but big change because we use this interface in many places.
Have you thought about just adding bool field to `Library`?
something like that:
```
type Library struct {
	main/project/... bool // need a good name that won't confuse
	ID                 string `json:",omitempty"`
	Name               string
	Version            string
	Indirect           bool          `json:",omitempty"`
	License            string        `json:",omitempty"`
	ExternalReferences []ExternalRef `json:",omitempty"`
	Locations          []Location    `json:",omitempty"`
}
```
But in this case we need to check all libs to find project. Perhaps it will spend more resurses.

knqyf263:
It also works for me.

DmitriyLewen:
@knqyf263  I updated logic. Take a look, please.

Also i added required changes in Trivy - https://github.com/aquasecurity/trivy/pull/3757/commits/f380a220021ea6d638183f41963ddd485a1d8aa8

knqyf263 (approved):

PR #196, opened 2023-03-17
fix(npm): improvement logic for lockfile v2-v3

DmitriyLewen:
## Description
1) npm includes optional dependencies in dependency list, graph and `node_modules` folder. We also need to include them.
    e.g.
    graph:
    ```
    3840@ /home/dmitriy/work/temp/3840
    └─┬ cra-append-sw@2.7.0
      └─┬ webpack@4.46.0
        └─┬ watchpack@1.7.5
          ├─┬ chokidar@3.5.3
          │ └── glob-parent@5.1.2
          └─┬ watchpack-chokidar2@2.0.1
            └─┬ chokidar@2.1.8
              └── glob-parent@3.1.0
    ```
    package-lock.json file:
    ```
    ...
        "node_modules/watchpack": {
          "version": "1.7.5",
          "resolved": "https://registry.npmjs.org/watchpack/-/watchpack-1.7.5.tgz",
          "integrity": "sha512-9P3MWk6SrKjHsGkLT2KHXdQ/9SNkyoJbabxnKOoJepsvJjJG8uYTR3yTPxPQvNDI3w4Nz1xnE0TLHK4RIVe/MQ==",
          "dependencies": {
            "graceful-fs": "^4.1.2",
            "neo-async": "^2.5.0"
          },
          "optionalDependencies": {
            "chokidar": "^3.4.1",
            "watchpack-chokidar2": "^2.0.1"
          }
        },
    ...
    ```
    folders:
    ```
    ➜  3840 ls -hl ./node_modules | grep watchpack
    drwxrwxr-x  3 dmitriy dmitriy 4,0K мар 15 11:41 watchpack
    
    ```
2. there are dependencies with `/` in package name.
e.g. [@babel/plugin-transform-classes](https://www.npmjs.com/package/@babel/plugin-transform-classes)
these dependencies start from `@`. We need to parse them properly.


pkg/nodejs/npm/parse.go
```diff
@@ -117,8 +119,13 @@ func (p *Parser) parseV2(packages map[string]Package) ([]types.Library, []types.
 		}
 		libs[pkgID] = lib
 
-		dependsOn := make([]string, 0, len(pkg.Dependencies))
-		for depName, depVersion := range pkg.Dependencies {
+		// npm builds graph using optional deps. e.g.:
+		// └─┬ watchpack@1.7.5
+		// ├─┬ chokidar@3.5.3 - optional dependency
+		// │ └── glob-parent@5.1.
+		dependencies := lo.Assign(pkg.Dependencies, pkg.OptionalDependencies)
```
knqyf263, line 126: Did you add a test for optional dependencies?

pkg/nodejs/npm/parse.go
```diff
@@ -253,6 +260,11 @@ func pkgNameFromPath(path string) string {
 	// node_modules/string-width
 	// node_modules/string-width/node_modules/strip-ansi
 	paths := strings.Split(path, "/")
+	// deps starting from `@` have pgkName with one `/`
+	// e.g. path == `node_modules/@babel/plugin-transform-classes` => pkgName == `@babel/plugin-transform-classes`
+	if strings.HasPrefix(paths[len(paths)-2], "@") {
```
knqyf263, line 265: It should not happen, but I think we should check the length just in case.

```suggestion
	if len(paths) >= 2 && strings.HasPrefix(paths[len(paths)-2], "@") {
```

pkg/nodejs/npm/parse.go
knqyf263: We can update the existing test case.

pkg/nodejs/npm/parse.go
DmitriyLewen: Done

pkg/nodejs/npm/parse.go
DmitriyLewen: Added optional deps to testcases

knqyf263 (approved):

PR #197, opened 2023-03-20
feat(php) add support for location, licenses and graph for composer.lock files

DmitriyLewen:
## Description
Add support for location, licenses and graph for `composer.lock` files.

`composer.lock` file doesn't contain info about Direct dependencies. 
We can parse it alongside `composer.json` file to find them.


pkg/php/composer/parse.go
```diff
@@ -24,18 +34,75 @@ func NewParser() types.Parser {
...
+			License:   strings.Join(pkg.License, ", "),
+			Locations: []types.Location{{StartLine: pkg.StartLine, EndLine: pkg.EndLine}},
+		}
+		libs[lib.Name] = lib
+
+		var dependsOn []string
+		for depName := range pkg.Require {
+			// Require field includes required php version, skip this
+			// Also skip PHP extensions
+			if depName != "php" || strings.HasPrefix(depName, "ext") {
```
knqyf263, line 62: I think it never reaches `strings.HasPrefix(depName, "ext")` due to `depName != "php"`. The comment says it skips PHP extensions, so I guess it should be as below.

```suggestion
			if depName != "php" && !strings.HasPrefix(depName, "ext") {
```

I prefer `A || B` rather than `!A && !B` to avoid this kind of bug, though.

```suggestion
			if depName == "php" || strings.HasPrefix(depName, "ext") {
				continue
			}
```

pkg/php/composer/parse.go
```diff
@@ -24,18 +34,75 @@ func NewParser() types.Parser {
...
+		}
+		dependsOn = sortDependsOn(dependsOn)
+		deps = append(deps, types.Dependency{
+			ID:        libID,
+			DependsOn: dependsOn,
 		})
 	}
-	return libs, nil, nil
+
+	return maps.Values(libs), deps, nil
```
knqyf263, line 89: Do we also want to sort libs?

DmitriyLewen: it makes sense.
I have moved the sort functions to the `utils` package. So, we can use these functions in other packages.

DmitriyLewen:
@knqyf263 i created changes after your notes. 
Also i need your opinion:
i added logic to parse `composer.json` file to detect direct dependencies - https://github.com/aquasecurity/trivy/pull/3873/files#diff-1956b9fdfc71894e0de000aa9ab290854ca7a967e3b44b7f25b335ab4a0a5730R130-R148
Do we need to move this logic in `go-dep-parser`? (i can do it in this PR).

knqyf263:
We can keep it in Trivy for now.

pkg/utils/utils.go
```diff
@@ -55,3 +55,24 @@ func MergeMaps(parent, child map[string]string) map[string]string {
 func PackageID(name, version string) string {
 	return fmt.Sprintf("%s@%s", name, version)
 }
+
+func SortLibs(libs []types.Library) []types.Library {
```
knqyf263: We can use sort.Interface

pkg/utils/utils.go
```diff
@@ -55,3 +55,24 @@ func MergeMaps(parent, child map[string]string) map[string]string {
...
+}
+
+func SortDeps(deps []types.Dependency) []types.Dependency {
+	sort.Slice(deps, func(i, j int) bool {
+		return deps[i].ID < deps[j].ID
+	})
+	return deps
+}
+
+func SortDependsOn(dependsOn []string) []string {
```
knqyf263: We can use sort.Strings

pkg/php/composer/parse.go
knqyf263: It is better to use sort.Interface. I've updated in https://github.com/aquasecurity/go-dep-parser/pull/197/commits/875540ff0617f5dc56f6567ef13dcde4c22412f5

knqyf263 (approved):

PR #199, opened 2023-03-24
Use uppercase Microsoft to fix building issues on case sensitive file systems

relaxnow:


relaxnow:
For posterity, go-rustaudit uses lowercase: https://github.com/aquasecurity/go-dep-parser/blob/main/go.mod#L1
Other modules like go-winio use uppercase: https://github.com/microsoft/go-winio/blob/main/go.mod#L1
This is a problem for case sensitive filesystems (like ext on Linux) when collaborating using the `vendor` folder because vendoring does not work case-insensitive: https://github.com/golang/go/issues/38342 . So if 1 person is on OS X and another on Linux vendoring will work differently for both and break (at least for the Linux person).

knqyf263:
Yes, I'm aware of the problem in general. But in this case, we use lowercase. Both Linux and macOS will use lowercase for go-rustaudit, doesn't it?

relaxnow:
No, I'm working with an app that's generating a single 'Microsoft' directory and throwing all the dependencies in it (including go-rustaudit). Oddly, I checked and this occurs even on a Linux builder.

knqyf263:
Sounds like capitalizing Microsoft doesn't address the issue. It may create a 'microsoft' directory and put all the dependencies in it. It will cause the same issue the other way around.

relaxnow:
Yes, certainly, it's a bit of a thorny issue that really should be fixed in Go but until it's fixed we're stuck with it.
In this particular project `go-dep-parser` and `syft` are the only ones using the proper lowercase name, which is what initially led me here. But that's a bit shortsighted. Again apologies.

PR #205, opened 2023-04-11
fix(java): fix overwriting project props in pom.xml

DmitriyLewen:
## Description
We don't need to overwrite project props (like `project.version`) from parent pom.xml files.
Some examples with `project.version` field:
1) `base` has `version` field => use value from this field for dependency(`example-api` == `1.0.0`):
image

2) `base` doesn't have `version` field, but `parent` and `top-parent` have `version` field.
=> use value from `parent`(`example-api` == `2.0.0`):
image
We currently **mistakenly** use `version` from `top-parent` (`example-api` == `3.0.0`).

3) when only `top-parent` has `version` field - use this value.

## Related Issues
- aquasecurity/trivy/issues/3986


knqyf263:
If I understand correctly, all we have to do is export inherited elements into project properties. I've added a commit. But I guess you will find more edge cases that my implementation will not work 🤣

knqyf263:
@yinonsh-orca Dmitriy is fixing another critical issue. He'll be back to this PR in a few days.

DmitriyLewen:
Hello @knqyf263 
Thanks for your refactoring! It looks better. 
I've played around with these changes and found no issues. It works as expected.

knqyf263 (approved):

PR #206, opened 2023-04-12
fix(yarn): ignore git protocol

DmitriyLewen:
## Description
Ignore `git` protocol.

## Related Issues
- aquasecurity/trivy/issues/4041


pkg/nodejs/yarn/parse_test.go
```diff
@@ -74,8 +74,8 @@ func TestParsePattern(t *testing.T) {
 		{
 			target:         `"jquery@git+https://xxxx:x-oauth-basic@github.com/tomoyamachi/jquery":`,
 			expectName:     "jquery",
-			expectProtocol: "git+https",
-			expactVersion:  "//xxxx:x-oauth-basic@github.com/tomoyamachi/jquery",
+			expectProtocol: "git",
```
knqyf263, line 77: I feel like the protocol should be `git+https` as documented here.
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#git-urls-as-dependencies

DmitriyLewen: Thanks! Fixed in c93968f35b4b339524aa15fa8ed348787167a799

knqyf263 (approved):

PR #207, opened 2023-04-13
refactor: use warning for invalid protocols in yarn.lock

DmitriyLewen:
## Description
We currently stop parsing if invalid protocol is found.
We need to show warning and continue parsing in these cases.

### DIFF
--- a/pkg/nodejs/yarn/parse.go
+++ b/pkg/nodejs/yarn/parse.go
@@ -208,7 +208,7 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine 
 				if !ignoreProtocol(protocol) {
 					// we need to calculate the last line of the block in order to correctly determine the line numbers of the next blocks
 					// store the error. we will handle it later
-					err = xerrors.Errorf("failed to parse package pattern: '%s', unknown protocol: '%s'", line, protocol)
+					err = xerrors.Errorf("unknown protocol: '%s', line: %s", protocol, line)
 					continue
 				}
 				continue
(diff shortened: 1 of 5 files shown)


pkg/nodejs/yarn/parse.go
```diff
@@ -216,6 +220,13 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine
 		}
 	}
 
+	// in case an unsupported protocol is detected
+	// show warning and continue parsing
+	if err != nil {
+		log.Logger.Warnf(err.Error())
```
knqyf263, line 226:
```suggestion
		log.Logger.Warnf("Yarn protocol error: %s", err)
```

pkg/nodejs/yarn/parse.go
```diff
@@ -205,7 +206,10 @@ func parseBlock(block []byte, lineNum int) (lib Library, deps []string, newLine
 			if patterns == nil || !validProtocol(protocol) {
 				skipBlock = true
 				if !ignoreProtocol(protocol) {
-					return Library{}, nil, -1, xerrors.Errorf("failed to parse package pattern: '%s', unknown protocol: '%s'", line, protocol)
+					// we need to calculate the last line of the block in order to correctly determine the line numbers of the next blocks
+					// store the error. we will handle it later
+					err = xerrors.Errorf("failed to parse package pattern: '%s', unknown protocol: '%s'", line, protocol)
```
knqyf263, line 211:
```suggestion
					err = xerrors.Errorf("unknown protocol: '%s', line: %s", protocol, line)
```

DmitriyLewen: Done

pkg/nodejs/yarn/parse.go
DmitriyLewen: Done

PR #209, opened 2023-04-19
fix(npm): add support of local packages for package-lock.json

DmitriyLewen:
## Description
For local packages(for example when `workspace` is used) `npm` uses links.(https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#packages)
e.g.:
```
    "functions/func": {
      "version": "1.0.0",
      ...
    },
    "node_modules/func": {
      "resolved": "functions/func",
      "link": true
    }
```
We need to correct find package name and dependsOn.

See [this test file](https://github.com/aquasecurity/go-dep-parser/pull/209/files#diff-163819972a365cb22abb0127ed1d8008b39399e58fe64978438e4704eb2f9054) to better understand this case.

This case is only for lockfile `v2` and `v3`.
It is not possible to determine local package version for lockfile `v1` because lock file doesn't have version for this package.
e.g.:
```
    "nested_func": {
      "version": "file:functions/nested_func",
    ...
    }

```
## Related Issues
- aquasecurity/trivy/issues/4039


knqyf263:
@nikpivkin has knowledge of workspaces. I'd defer to him.

pkg/nodejs/npm/parse.go
```diff
@@ -167,6 +205,26 @@ func findDependsOn(pkgPath, depName string, packages map[string]Package) (string
 			return utils.PackageID(depName, dep.Version), nil
 		}
 	}
+
+	// There are cases when pkgPath doesn't have `node_modules` suffix (e.g. when it is local package)
+	// add `node_modules` and try to find dep
+	// e.g. for pkgPath == `function/func1`, depName == "debug`:
+	// `node_modules/debug`
+	// case when `function/func1/node_modules_debug` exists resolved in loop
```
nikpivkin, line 213: should there be a `function/func1/node_modules/debug` here?

DmitriyLewen: Right! It is typo.
I fixed this in https://github.com/aquasecurity/go-dep-parser/pull/209/commits/57989a6aa3f30ee56418ddffec957f42c438b68c

nikpivkin (approved):
lgtm

knqyf263 (approved):

PR #210, opened 2023-04-20
fix(pom): skipping missing modules

liorj-orca:
Following the issue described at https://github.com/aquasecurity/trivy/issues/3747 where for the case where you run :
trivy fs pom.xml 
with the following pom.xml:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <dependencies>
        <dependency>
            <groupId>org.apache.hive</groupId>
            <artifactId>hive-exec</artifactId>
            <version>2.3.3</version>
        </dependency>
    </dependencies>
</project>
```

you get zero results due to a parser error of one of the dependencies in the pom.
The following PR suggests skipping those missing modules.

### DIFF
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -156,7 +156,8 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency, 
 		for _, relativePath := range result.modules {
 			moduleArtifact, err := p.parseModule(result.filePath, relativePath)
 			if err != nil {
-				return nil, nil, xerrors.Errorf("module error (%s): %w", relativePath, err)
+				log.Logger.Debugf("Unable to parse %q module: %s", result.filePath, err)
+				continue
 			}
 
 			queue.enqueue(moduleArtifact)
--- a/pkg/java/pom/parse_test.go
+++ b/pkg/java/pom/parse_test.go
@@ -442,10 +442,15 @@ func TestPom_Parse(t *testing.T) {
 			},
 		},
 		{
-			name:      "module not found",
+			name:      "module not found - unable to parse module",
 			inputFile: filepath.Join("testdata", "not-found-module", "pom.xml"),
 			local:     true,
-			wantErr:   "stat testdata/not-found-module/module: no such file or directory",
+			want: []types.Library{
+				{
+					Name:    "com.example:aggregation",
+					Version: "1.0.0",
+				},
+			},
 		},
 	}
 	for _, tt := range tests {


liorj-orca:
@knqyf263 could you review this please?

pkg/java/pom/parse.go
```diff
@@ -156,7 +156,9 @@ func (p *parser) parseRoot(root artifact) ([]types.Library, []types.Dependency,
 		for _, relativePath := range result.modules {
 			moduleArtifact, err := p.parseModule(result.filePath, relativePath)
 			if err != nil {
-				return nil, nil, xerrors.Errorf("module error (%s): %w", relativePath, err)
+				log.Logger.Error(xerrors.Errorf("module error (%s): %w - Some modules may be missing. "+
+					"Skipping to the next module", relativePath, err))
```
DmitriyLewen, line 160: We usually use `Debug` for wrong part of dependency. What do you think about this:
```suggestion
				log.Logger.Debugf("Unable to parse %q module: %s", result.filePath, err)
```

DmitriyLewen (commented):
Hello @liorj-orca 
Thanks for your work!

I added 1 comment.

Can you also add test for this case?

Regards, Dmitriy

liorj-orca:
Hi @DmitriyLewen, I changed the log level to debug.
regarding the test, I noticed there was already a test with the same "issue" that I had to change its expected output - let me know if you would like me to add another dedicated test specifically for that as I believe the test now covers both cases.

DmitriyLewen:
Hello @liorj-orca 
Thanks for you work!

> let me know if you would like me to add another dedicated test specifically for that as I believe the test now covers both cases.

Good solution!
I think that's enough.

PR currently looks good to me.

DmitriyLewen (approved):

liorj-orca:
@DmitriyLewen what else needs to be done so we could merge this and update trivy go.mod so the change would be populated?

DmitriyLewen:
Hello @liorj-orca 
We then need @knqyf263 to look at that PR and merge it.

knqyf263 (approved):

DmitriyLewen:
Hello @liorj-orca 
We merged this PR.

Can you open new PR in Trivy and update `go-dep-parser` version?
Some recomendation about PRs [here](https://aquasecurity.github.io/trivy/v0.41/community/contribute/pr/).

PR #211, opened 2023-04-21
fix(pip): remove extras from lines of requirement.txt files

DmitriyLewen:
## Description
`requirement.txt` can contain extras. e.g. `pyjwt[crypto]==2.1.0`
We need to find dependency name correctly.

### DIFF
--- a/pkg/python/pip/parse.go
+++ b/pkg/python/pip/parse.go
@@ -14,6 +14,8 @@ const (
 	commentMarker string = "#"
 	endColon      string = ";"
 	hashMarker    string = "--"
+	startExtras   string = "["
+	endExtras     string = "]"
 )
 
 type Parser struct{}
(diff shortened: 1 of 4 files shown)


pkg/python/pip/parse.go
```diff
@@ -54,3 +58,13 @@ func rStripByKey(line string, key string) string {
 	}
 	return line
 }
+
+func removeExtras(line string) string {
+	startIndex := strings.Index(line, startExtras)
+	endIndex := strings.Index(line, endExtras) + 1
+	if startIndex != -1 && endIndex != -1 {
+		log.Logger.Debugf("unable to get name and version for extras of %q. Run `pip3 freeze > requirements.txt` to update file and get this dependency", line)
```
knqyf263, line 66: I don't think we need this message.

```
setup(
    name="example",
    ...
    install_requires=[
        "requests"],
    extras_require={
        'test':  ["nose"],
        'doc': ["sphinx"],
    }
)```

In the above example, the following commands download different dependencies.

```
pip install example[test] # install request and nose
pip install example[doc] # install request and sphinx
pip install example # install request
```

Even when extras are not specified, we can't get the names and versions of indirect dependencies (`requiest` in this case). So, recommending `pip freeze` in our doc is enough.

DmitriyLewen: okay, i removed it in https://github.com/aquasecurity/go-dep-parser/pull/211/commits/47b299f05c1769ae4d62c211f1bbe2626fac9530

knqyf263 (approved):

PR #217, opened 2023-05-16
feat(python): add license fields

afdesk:
`License` field is deprecated and replaced by the new `License-Expression` field, and there was added `License-File` field.
Read more here [PEP 639 – Improving License Clarity with Better Package ](https://peps.python.org/pep-0639/#deprecate-license-field).

also there are legacy packages that use license classifiers (`Classifier` field), so we have to check it too.

### DIFF
--- a/pkg/python/packaging/parse.go
+++ b/pkg/python/packaging/parse.go
@@ -36,6 +36,7 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
 			if strings.HasPrefix(classifier, "License ::") {
 				if values := strings.Split(classifier, " :: "); len(values) > 1 {
 					license = values[len(values)-1]
+					break
 				}
 			}
 		}
--- a/pkg/python/packaging/parse.go
+++ b/pkg/python/packaging/parse.go
@@ -33,11 +33,10 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
 	}
 	if license == "" {
 		for _, classifier := range h.Values("Classifier") {
-			if strings.HasPrefix(classifier, "License ::") {
-				if values := strings.Split(classifier, " :: "); len(values) > 1 {
-					license = values[len(values)-1]
-					break
-				}
+			if strings.HasPrefix(classifier, "License :: ") {
+				values := strings.Split(classifier, " :: ")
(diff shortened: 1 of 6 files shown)


pkg/python/packaging/parse.go
```diff
@@ -26,11 +27,26 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
...
+	}
+	if license == "" {
+		for _, classifier := range h.Values("Classifier") {
+			if strings.HasPrefix(classifier, "License :: ") {
+				values := strings.Split(classifier, " :: ")
+				license = values[len(values)-1]
+				break
+			}
+		}
+	}
```
knqyf263, line 42: We may want to flatten the conditions.

```suggestion
	// "License-Expression" takes precedence as "License" is deprecated.
	// cf. https://peps.python.org/pep-0639/#deprecate-license-field
	var license string
	if l := h.Get("License-Expression"); l != "" {
		license = l
	else if l := h.Get("License"); l != "" {
		license = l
	else {
		for _, classifier := range h.Values("Classifier") {
			if strings.HasPrefix(classifier, "License :: ") {
				values := strings.Split(classifier, " :: ")
				license = values[len(values)-1]
				break
			}
		}
	}
```

pkg/python/packaging/parse.go
```diff
@@ -26,11 +27,26 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
...
+
 	return []types.Library{
 		{
-			Name:    h.Get("Name"),
-			Version: h.Get("Version"),
-			License: h.Get("License"),
+			Name:        h.Get("Name"),
+			Version:     h.Get("Version"),
+			License:     license,
+			LicenseFile: h.Get("License-File"),
```
knqyf263, line 49: It could be a bit hacky, but I'm not keen on adding a new field only for python. What if adding `fil://license.txt` to `License` if the license is empty?

afdesk: let's try this way.

pkg/python/packaging/parse.go
afdesk: ok, sure

afdesk:
@knqyf263 could you take a look at this PR again?
@nikpivkin will finish this task.

PR #219, opened 2023-05-22
feat: implement Julia parser

Octogonapus:
This PR implements parsing for Julia Manifests.

I'm adding this as part of https://github.com/aquasecurity/trivy/pull/4438


Octogonapus:
@knqyf263 This is RTM from my end

Octogonapus:
@knqyf263 Do I need to do anything else on this PR?

Octogonapus:
@knqyf263 @simar7 Any chance this will ever get reviewed or merged?

knqyf263:
@DmitriyLewen Can you please take a look first?

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// We can't know the Julia version on an old manifest.
+	// All newer manifests include a manifest version and a julia version.
+	if man.ManifestFormat == "" {
+		man = Manifest{
+			JuliaVersion:   "unknown",
+			ManifestFormat: "unknown",
+			Dependencies:   oldDeps,
+		}
+	}
```
DmitriyLewen, line 52: If i understand correctly - there are 2 manifest format for us:
- before 1.8
- after 1.8

But we don't use manifest format for parsing. So we can just compare with `nil` and remove `ManifestFormat`.
```suggestion
	if man.ManifestFormat == nil {
		man = Manifest{
			JuliaVersion:   "unknown",
			Dependencies:   oldDeps,
		}
	}
```

Correct me, if i have missed something.

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,118 @@
...
+	var deps []types.Dependency
+	for name, manifestDeps := range man.Dependencies {
+		if len(manifestDeps) > 1 {
+			return nil, nil, xerrors.Errorf("multiple entries for dep: %s", name)
+		}
+		manifestDep := manifestDeps[0]
+		version := depVersion(manifestDep, man.JuliaVersion)
+		pkgID := utils.PackageID(manifestDep.UUID, version)
+		lib := types.Library{
+			ID:      pkgID,
```
DmitriyLewen, line 72: If i understand correctly - we can use `uuid` for `ID` field - https://pkgdocs.julialang.org/v1/toml-files/#Multiple-packages-with-the-same-name

Can you also add test for case when there are 2 dependencies with same names, but different versions and uuid.

pkg/julia/manifest/parse_test.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f, err := os.Open(tt.file)
+			require.NoError(t, err)
+
+			got, deps, err := NewParser().Parse(f)
+			require.NoError(t, err)
+
+			sortLibs(got)
+			sortLibs(tt.want)
```
DmitriyLewen, line 63: use already sorted Libs:
https://github.com/aquasecurity/go-dep-parser/blob/8dd9c9c78a18c7829657cc3c61c168981661fa16/pkg/julia/manifest/parse.go#L86-L87

pkg/julia/manifest/parse_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			assert.Equal(t, tt.want, got)
+			if tt.wantDeps != nil {
+				sortDeps(deps)
+				sortDeps(tt.wantDeps)
```
DmitriyLewen, line 68: same

pkg/julia/manifest/parse_test.go
```diff
@@ -0,0 +1,98 @@
...
+func TestParse(t *testing.T) {
+	tests := []struct {
+		name     string
+		file     string // Test input file
+		want     []types.Library
+		wantDeps []types.Dependency
+	}{
+		{
+			name:     "Manifest v1.6",
+			file:     "testdata/primary/Manifest_v1.6.toml",
```
DmitriyLewen, line 24: You only use `Manifest.toml` files. 
Remove other files ( like `Project.toml`, `README.md`, etc...) files, please.

pkg/julia/manifest/parse_test.go
```diff
@@ -0,0 +1,98 @@
...
+			wantDeps: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			f, err := os.Open(tt.file)
+			require.NoError(t, err)
+
+			got, deps, err := NewParser().Parse(f)
```
DmitriyLewen, line 59:
```suggestion
			gotLibs, gotDeps, err := NewParser().Parse(f)
```

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,118 @@
...
+			deps = append(deps, *dep)
+		}
+	}
+	sort.Sort(types.Libraries(libs))
+	sort.Sort(types.Dependencies(deps))
+	return libs, deps, nil
+}
+
+// Returns the dependencies in `deps` of the given `dep`. If there are no dependencies, returns `nil`.
+func parseDependencies(pkgId string, dep Dependency, deps map[string][]Dependency, juliaVersion string) *types.Dependency {
```
DmitriyLewen, line 92: These names can confusing.
What if use something like this:
```suggestion
func parseDependencies(pkgId string, pkgDeps []string, allDeps map[string][]Dependency, juliaVersion string) *types.Dependency {
```

DmitriyLewen (commented):
Hello @Octogonapus 
Thanks for your work!

I left some comments.
Take a look, please.

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,118 @@
...
+	pkgParser := naivePkgParser{r: r}
+	lineNumIdx := pkgParser.parse()
+
+	var libs []types.Library
+	var deps []types.Dependency
+	for name, manifestDeps := range man.Dependencies {
+		if len(manifestDeps) > 1 {
+			return nil, nil, xerrors.Errorf("multiple entries for dep: %s", name)
+		}
+		manifestDep := manifestDeps[0]
```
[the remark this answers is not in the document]
Octogonapus: Added

pkg/julia/manifest/parse_test.go
Octogonapus: Removed

pkg/julia/manifest/parse.go
Octogonapus: Done

pkg/julia/manifest/parse.go
Octogonapus: Done - required a small rewrite

Octogonapus:
Thank you for looking at this. Review comments have been addressed.

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,118 @@
...
+	return &Parser{}
+}
+
+func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var oldDeps map[string][]Dependency
+	var man Manifest
+	decoder := toml.NewDecoder(r)
+	// Try to read the old Manifest format. If that fails, try the new format.
+	if _, err := decoder.Decode(&oldDeps); err != nil {
+		r.Seek(0, io.SeekStart)
```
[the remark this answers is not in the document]
DmitriyLewen:
```suggestion
		if _, err := r.Seek(0, io.SeekStart); err != nil {
		     return nil, nil, xerrors.Errorf("seek error: %w", err)
	}
```

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,218 @@
...
+		finalDeps := make(map[string]*string)
+		for _, depName := range possibleDeps {
+			finalDeps[depName] = nil
+		}
+		return &DecodedDependency{
+			Dependencies: finalDeps,
+			UUID:         dep.UUID,
+			Version:      dep.Version,
+		}, nil
+	}
```
DmitriyLewen, line 201: I am not sure that we need both `PrimitiveDependency` and `DecodedDependency`.
What if we will add :
```
DependsOs    []string       `toml:"-"`       // list of dependent UUID's.
``` 
 field here - https://github.com/aquasecurity/go-dep-parser/blob/275db58029ca208af1928c0ab925929e37804444/pkg/julia/manifest/parse.go#L27-L31
and handle uuids for `["A", "B"]` case in this function.
Something like that:
```suggestion
// Decodes a primitive dependency using the metadata from parse time.
func decodeDependency(man *PrimitiveManifest, dep *PrimitiveDependency, metadata *toml.MetaData) (*DecodedDependency, error) {
	// Try to decode as []string first where the manifest looks like deps = ["A", "B"]
	var possibleDeps []string
	err := metadata.PrimitiveDecode(dep.Dependencies, &possibleDeps)
	if err == nil {
		var possibleUuids []string
		for _, depName := range possibleDeps {
			primDep := man.Dependencies[depName]
			if len(primDep) > 1 {
				// it is case for [deps.A.deps]
				// return error
			}
			possibleUuids = append(possibleUuids, primDep[0].UUID)
		}
		dep.DependsOs = possibleUuids
		return dep, nil
	}
```
`[deps.A.deps]` case already uses UUID

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,218 @@
...
+	// naive parser to get line numbers
+	pkgParser := naivePkgParser{r: r}
+	lineNumIdx := pkgParser.parse()
+
+	var libs []types.Library
+	var deps []types.Dependency
+	for name, manifestDeps := range man.Dependencies {
+		for _, manifestDep := range manifestDeps {
+			version := depVersion(&manifestDep, man.JuliaVersion)
+			pkgID := utils.PackageID(manifestDep.UUID, version)
```
DmitriyLewen, line 85: pkgID still contains version.

DmitriyLewen (commented):
@Octogonapus left some idea. Can you take a look?

pkg/julia/manifest/parse.go
Octogonapus: That definition of `PrimitiveDependency` can't get parsed: `toml: line 10 (last key "deps.A.deps"): incompatible types: TOML value has type map[string]interface {}; destination has type slice`. That error is why I delayed the parsing until later in `decodeDependency`.

pkg/julia/manifest/parse.go
Octogonapus: Removed

pkg/julia/manifest/parse.go
DmitriyLewen: Perhaps i confused you.
I meant that we stay your solution for `PrimitiveDependency`, but add 1 more field - https://gist.github.com/DmitriyLewen/22431f167fdb110b5b17a90b6071051b#file-parse-go-L28-L33

Also we need to use `PrimitiveDecode` twice ( as you did that):
https://gist.github.com/DmitriyLewen/22431f167fdb110b5b17a90b6071051b#file-parse-go-L140-L172

My code in gist needs refactoring, but i hope you will understand my logic.
PS. sorry, i didn't pull your changes for ID.

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// The other possibility is a map where the manifest looks like
+	// [deps.A.deps]
+	// B = "..."
+	var possibleDepsMap map[string]string
+	err = metadata.PrimitiveDecode(dep.Dependencies, &possibleDepsMap)
+	if err == nil {
+		possibleUuids := maps.Values(possibleDepsMap)
+		dep.DependsOn = possibleUuids
```
DmitriyLewen, line 153: If i remember correctly - `maps.Values` returns values in random order. Can you add sort here?

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,159 @@
...
+	if err == nil {
+		var possibleUuids []string
+		for _, depName := range possibleDeps {
+			primDep := man.Dependencies[depName]
+			if len(primDep) > 1 {
+				return primitiveDependency{}, xerrors.Errorf("Dependency %q has invalid format (parsed multiple deps): %s", depName, primDep)
+			}
+			possibleUuids = append(possibleUuids, primDep[0].UUID)
+		}
+		dep.DependsOn = possibleUuids
```
DmitriyLewen, line 142: I think we can add sort UUIDs here

pkg/julia/manifest/parse.go
```diff
@@ -0,0 +1,159 @@
...
+	return &Parser{}
+}
+
+func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var oldDeps map[string][]primitiveDependency
+	var primMan primitiveManifest
+	var manMetadata toml.MetaData
+	decoder := toml.NewDecoder(r)
+	// Try to read the old Manifest format. If that fails, try the new format.
+	if _, err := decoder.Decode(&oldDeps); err != nil {
```
DmitriyLewen, line 40: Looks like we need to return `manMetadata` here too.

DmitriyLewen (commented):
left 3 small comments

pkg/julia/manifest/parse.go
Octogonapus: Done

pkg/julia/manifest/parse.go
Octogonapus: Done

pkg/julia/manifest/parse.go
Octogonapus: No need, because the old deps format can always be parsed into that object. We only need the parse metadata for the new manifest format.

DmitriyLewen: okay, thanks for this explain!

DmitriyLewen (approved):
Thank you for this work!
Now this PR seems good to me.

@knqyf263 can you take a look and merge this PR, if it is okay for you?

Octogonapus:
@DmitriyLewen @knqyf263 bump

Octogonapus:
@knqyf263 @DmitriyLewen bump, what else is needed for merge?

DmitriyLewen:
Hello @Octogonapus 
I don't have permission to merge this PR.

@knqyf263 has some priority tasks. 
Give us a little more time, please.

knqyf263:
We've not seen any demand for [this feature](https://github.com/aquasecurity/trivy/discussions/4437) so far. Once this PR gets merged, maintainers need to keep maintaining. We probably want to suspend this task until the community needs it.

Octogonapus:
@knqyf263 I understand. In the meantime, I asked the Julia community if others wanted this feature. They have added their support to https://github.com/aquasecurity/trivy/discussions/4437

knqyf263:
>@knqyf263 I understand. In the meantime, I asked the Julia community if others wanted this feature. They have added their support to https://github.com/aquasecurity/trivy/discussions/4437

OK, so let's merge your PR! But as we're not so familiar with Julia, we probably need support from the community.

DmitriyLewen:
Hello @Octogonapus 
Can you reopen and update you PR for Trivy now?

Octogonapus:
Thank you!

> we probably need support from the community

Yes I can help maintain this.

> Can you reopen and update you PR for Trivy now?

Yes I will.

PR #220, opened 2023-05-24
fix(pnpm): change logic to parse package name and version

DmitriyLewen:
## Description
We have bug with packages from registry (see https://github.com/aquasecurity/trivy/issues/3793#issuecomment-1460485444).
Also `pnpm` released lockfile `v6`. 
Change logic for both cases.


pkg/nodejs/pnpm/parse.go
```diff
@@ -100,10 +100,73 @@ func isIndirectLib(name string, directDeps map[string]string) bool {
...
+// relative path: `/<pkg_name>@<pkg_version>` e.g. /foo@1.0.0
+// registry: `<registry_url>/<pkg_name>@<pkg_version>` e.g. registry.node-modules.io/foo@1.0.0
+// https://github.com/pnpm/pnpm/pull/5810
+// https://github.com/pnpm/spec/issues/4#issuecomment-1558891433
+//
+// `pkg_name` has 2 formats:
+// with slash - @<author>@<name>. e.g. `/@babel/generator@7.21.9`
+// without slash - <name>. e.g. `/lodash@4.17.10`
+//
+// `pkg_version` can contain peer deps. <pkg_name>@<pkg_version>(<peer_deps>)
```
knqyf263, line 149: I think we can simplify the implementation.
https://go.dev/play/p/KyThjdKYGRi

```suggestion
func parsePackage(pkg, versionSep string) (string, string) {
	// Skip registry
	// e.g.
	//    - "registry.npmjs.org/lodash/4.17.10" => "lodash/4.17.10"
	//    - "registry.npmjs.org/@babel/generator/7.21.9" => "@babel/generator/7.21.9"
	//    - "/lodash/4.17.10" => "lodash/4.17.10"
	_, pkg, _ = strings.Cut(pkg, "/")

	// Parse namespace(?)
	// e.g.
	//    - v5:  "@babel/generator/7.21.9" => {"babel", "generator/7.21.9"}
	//    - v6+: "@babel/helper-annotate-as-pure@7.18.6" => "{"babel", "helper-annotate-as-pure@7.18.6"}
	var namespace string
	if strings.HasPrefix(pkg, "@") {
		namespace, pkg, _ = strings.Cut(pkg, "/")
	}

	// Parse package name
	// e.g.
	//    - v5:  "generator/7.21.9" => {"generator", "7.21.9"}
	//    - v6+: "helper-annotate-as-pure@7.18.6" => {"helper-annotate-as-pure", "7.18.6"}
	var name, version string
	name, version, _ = strings.Cut(pkg, versionSep)

	if namespace != "" {
		name = fmt.Sprintf("%s/%s", namespace, name)
	}

	// Trim peer deps
	// e.g.
	//    - v5:  "7.21.5_@babel+core@7.21.8" => "7.21.5"
	//    - v6+: "7.21.5(@babel/core@7.20.7)" => "7.21.5"
	if idx := strings.IndexAny(version, "_("); idx != -1 {
		version = version[:idx]
	}

	return name, version
}
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -100,10 +100,73 @@ func isIndirectLib(name string, directDeps map[string]string) bool {
...
 
-func getPackageNameAndVersion(pkg string) (string, string) {
-	idx := strings.LastIndex(pkg, "/")
-	name := pkg[1:idx]
-	version := pkg[idx+1:]
+func getPackageNameAndVersion(pkg string, lockFileVersion int8) (string, string) {
+	if lockFileVersion < 6 {
+		return getPackageNameAndVersionV5(pkg)
+	}
+	return getPackageNameAndVersionV6(pkg)
```
knqyf263, line 107:
```suggestion
	versionSep := "@"
	if lockFileVersion < 6 {
		versionSep = "/"
	}
	return parsePackage(pkg, versionSep)
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -100,10 +100,73 @@ func isIndirectLib(name string, directDeps map[string]string) bool {
...
+}
+
+// 2 package formats are possible:
+// relative path: `/<pkg_name>/<pkg_version>` e.g. /foo/1.0.0
+// registry: `<registry_url>/<pkg_name>/<pkg_version>` e.g. registry.node-modules.io/foo/1.0.0
+// https://github.com/pnpm/spec/blob/master/lockfile/5.2.md#packages
```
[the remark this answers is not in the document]
DmitriyLewen: Done - https://github.com/aquasecurity/go-dep-parser/pull/220/commits/c7552c05e69ec1d04f56224668c977fe12190dc8

pkg/nodejs/pnpm/parse.go
DmitriyLewen: this way doesn't work with repository name with additional `/`. I mean `npm-proxy.fury.io/dmitriytest/lodash/5.0.0`.
I added test for that - https://github.com/aquasecurity/go-dep-parser/pull/220/files#diff-828b2b43cb2b887728250ad289067f4874a1e6d0a52fb6d339f1020c0f4c28ecR105-R111

knqyf263: I didn't see such a case in the tests, so I didn't expect the registry also contains a slash.

knqyf263: I read through pnpm logic. It doesn't allow more slashes in registries.
https://github.com/pnpm/dependency-path/blob/02815d16d05d513a5103feaeafdf88716cb72bce/src/index.ts#L61-L86

knqyf263: Looks like they adopt the same approach as mine.

knqyf263: This is probably a new one.
https://github.com/pnpm/pnpm/blob/ce61f8d3c29eee46cee38d56ced45aea8a439a53/lockfile/lockfile-file/src/experiments/inlineSpecifiersLockfileConverters.ts#L162-L167

knqyf263: My conclusion so far is below.

- registries don't allow `/`.
    - We don't need the test case like `npm-proxy.fury.io/dmitriytest/lodash/5.0.0`
- Our logic is missing file support.
    - https://github.com/pnpm/pnpm/blob/ce61f8d3c29eee46cee38d56ced45aea8a439a53/lockfile/lockfile-file/test/lockfileV6Converters.test.ts#L36
    - need semver check?
        - https://github.com/pnpm/pnpm/blob/ce61f8d3c29eee46cee38d56ced45aea8a439a53/packages/dependency-path/src/index.ts#LL148C6-L148C6

pkg/nodejs/pnpm/parse.go
DmitriyLewen: Done

pkg/nodejs/pnpm/parse.go
DmitriyLewen: > We don't need the test case like npm-proxy.fury.io/dmitriytest/lodash/5.0.0

I used you logic.

> Our logic is missing file support.
need semver check? 

Made in https://github.com/aquasecurity/go-dep-parser/pull/220/commits/6f9e3f0612b8ca5e2031a2e249d626bdd82f7c39

pkg/nodejs/pnpm/parse.go
```diff
@@ -64,13 +69,29 @@ func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency)
...
 
+	var lockVer float64
+	switch v := lockFile.LockfileVersion.(type) {
+	// lock
+	case float64:
+		lockVer = v
+	case string:
+		var err error
+		if lockVer, err = strconv.ParseFloat(v, 64); err != nil {
+			log.Logger.Debugf("unable to convert lock file version: %s", err)
```
knqyf263, line 80: Log message should be capitalized.

```suggestion
			log.Logger.Debugf("Unable to convert lock file version: %s", err)
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -64,13 +69,29 @@ func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency)
...
+			return nil, nil
+		}
+	default:
+		log.Logger.Debugf("unable to convert lock file version: %s", lockFile.LockfileVersion)
```
knqyf263, line 84:
```suggestion
		log.Logger.Debugf("Unable to convert lock file version: %s", lockFile.LockfileVersion)
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -95,15 +116,61 @@ func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency)
...
 	_, ok := directDeps[name]
 	return !ok
 }
 
-func getPackageNameAndVersion(pkg string) (string, string) {
-	idx := strings.LastIndex(pkg, "/")
-	name := pkg[1:idx]
-	version := pkg[idx+1:]
-
+func getPackageNameAndVersion(pkg, name, version string, lockFileVersion float64) (string, string) {
```
knqyf263, line 124: We should use the same terminology in pnpm.

```suggestion
func getPackageNameAndVersion(depPath, name, version string, lockFileVersion float64) (string, string) {
```

pkg/nodejs/pnpm/parse.go
```diff
@@ -95,15 +116,61 @@ func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency)
...
+	// local archives have `name` and `version` fields
+	// https://github.com/pnpm/spec/blob/ad27a225f81d9215becadfa540ef05fa4ad6dd60/lockfile/5.2.md#packagesdependencypathname
+	if strings.HasPrefix(pkg, "file:") {
```
knqyf263, line 127: I thought archives started with `@` like `@registry.npmjs.org/is-positive/-/is-positive-1.0.0.tgz`
https://github.com/pnpm/pnpm/blob/ce61f8d3c29eee46cee38d56ced45aea8a439a53/lockfile/lockfile-file/test/lockfileV6Converters.test.ts#LL164C7-L164C63

Where is the `file:` prefix defined?

pkg/nodejs/pnpm/parse.go
```diff
@@ -95,15 +116,61 @@ func (p *Parser) parse(lockFile *LockFile) ([]types.Library, []types.Dependency)
...
+		}
+		return name, version
+	}
+	versionSep := "@"
+	if lockFileVersion < 6 {
+		versionSep = "/"
+	}
+	return parsePackage(pkg, versionSep)
+}
+func parsePackage(pkg, versionSep string) (string, string) {
```
knqyf263, line 140:
```suggestion
// cf. https://github.com/pnpm/pnpm/blob/ce61f8d3c29eee46cee38d56ced45aea8a439a53/packages/dependency-path/src/index.ts#L112-L163
func parseDependencyPath(pkg, versionSep string) (string, string) {
```

pkg/nodejs/pnpm/parse.go
DmitriyLewen: there are 3 cases:
- local archive
- git
- remove archive

`pnpm` parses that [here](https://github.com/pnpm/pnpm/blob/04a279881a47b0d7b5a32a5bc774d0ff071d1e2f/fetching/pick-fetcher/src/index.ts#L7C1-L14)

There is also install from github. This is also archive,  but there is no information about it in the package name (only `resolution` field says that it is archive).:
```
  github.com/debug-js/debug/d1616622e4d404863c5a98443f755b4006e971dc:
    resolution: {tarball: https://codeload.github.com/debug-js/debug/tar.gz/d1616622e4d404863c5a98443f755b4006e971dc}
    name: debug
    version: 4.3.4
    engines: {node: '>=6.0'}
    peerDependencies:
      supports-color: '*'
    peerDependenciesMeta:
      supports-color:
        optional: true
    dependencies:
      ms: 2.1.2
    dev: false
```

all these cases (archives and github) have `name` and `version` fields.
Looks like it's wrong to parse  archive name, because archive can use any name (look this -https://github.com/aquasecurity/go-dep-parser/pull/220/files#diff-fc1d9989ef2210b568c552d504211fa3ff47735ea7b17865f3942921d516d89cR149 you can use any name, but `name` and `version` field will be correct.):
```
  file:foo/bar/lodashhhh-any-version.tgz:
    resolution: {integrity: sha512-x06dhjWXFr2GpUpY0LyzYOAUIppa/T/PpY4O7qu9kIAJESqDjOqTur552ggOOfaZNdAervnxmXpFJIRGULsl+w==, tarball: file:foo/bar/lodashhhh-any-version.tgz}
    name: lodash
    version: 4.17.10
    dev: false

```
I think we can use next logic:
1) check `name` and `version` fields (we will parse archives and packages installed from github)
2) if these fields are empty - use `getPackageNameAndVersion` function.

@knqyf263 wdyt?

pkg/nodejs/pnpm/parse.go
DmitriyLewen: Made in https://github.com/aquasecurity/go-dep-parser/pull/220/commits/2babc94911560f26355064c29a8340c4765ad992

pkg/nodejs/pnpm/parse.go
DmitriyLewen: Made in https://github.com/aquasecurity/go-dep-parser/pull/220/commits/2babc94911560f26355064c29a8340c4765ad992

pkg/nodejs/pnpm/parse.go
DmitriyLewen: Made in https://github.com/aquasecurity/go-dep-parser/pull/220/commits/2babc94911560f26355064c29a8340c4765ad992

pkg/nodejs/pnpm/parse.go
DmitriyLewen: Made in https://github.com/aquasecurity/go-dep-parser/pull/220/commits/627bec841fb40712e7e842380ef56b3153929377

pkg/nodejs/pnpm/parse.go
DmitriyLewen: we can find tarballs from `Resolution.Tarball` field. Take a look on this commit - https://github.com/aquasecurity/go-dep-parser/pull/220/commits/584f3c9a53e5ec77c04c9c54ad430482abc80361

knqyf263: Looks good.

knqyf263 (approved):

PR #221, opened 2023-05-30
fix(jar): remove duplicates of jar libs

DmitriyLewen:
## Description
See [#4496](aquasecurity/trivy/issues/4496)

## Related Issues
- aquasecurity/trivy/issues/4496


pkg/java/jar/parse.go
```diff
@@ -413,3 +417,19 @@ func (m manifest) determineVersion() (string, error) {
...
+		l, ok := uniqLibs[lib.Name]
+		// comparing Version and FilePath
+		if ok && lib.Version == l.Version && lib.FilePath == l.FilePath {
+			continue
+		}
+		uniqLibs[lib.Name] = lib
+	}
+	libSlice := maps.Values(uniqLibs)
+	sort.Sort(types.Libraries(libSlice))
+	return libSlice
```
knqyf263, line 434: What do you think about this approach?

```suggestion
	return := lo.FindDuplicatesBy(libs, func(lib types.Library)  {
			return fmt.Sprintf("%s::%s::%s", lib.Name, lib.Version, lib.FilePath)
	})
```

DmitriyLewen: `lo.FindDuplicatesBy` takes only duplicates and skips unique libs.
`lo.FindUniquesBy` takes only unique libs and skips duplicates.

`types.Library` is not comparable type, so we can't use `slice.Contains` function to find duplicates in `libs`.

But i found 1 way - we will take all uniq libs and one lib for each duplicate - https://github.com/aquasecurity/go-dep-parser/pull/221/commits/e0a855ff8ab75db4bd34290274aea25dbe7641c5
I added tests so you can see it.

DmitriyLewen: or we can use `map[string]types.Library` here - https://github.com/aquasecurity/go-dep-parser/blob/307b489b6ab80c6f2f2159a697f6c5489985635e/pkg/java/jar/parse.go#L91
and check for unique libs when append into this map.

DmitriyLewen: FYI [this](https://github.com/aquasecurity/go-dep-parser/pull/221/commits/e0a855ff8ab75db4bd34290274aea25dbe7641c5#diff-4191f5e37494ca6155f14396af0beb3c85994d231580ee4d58ccd3b691fe5696R331) and [this](https://github.com/aquasecurity/go-dep-parser/pull/221/commits/e0a855ff8ab75db4bd34290274aea25dbe7641c5#diff-4191f5e37494ca6155f14396af0beb3c85994d231580ee4d58ccd3b691fe5696R375) testscases are contained in `TestParse`.
I have updated  `io.quarkus.gizmo.gizmo-1.1.1.Final.jar` test file to move [testcase when jar contains 2 jars with same name and different versions](https://github.com/aquasecurity/go-dep-parser/pull/221/commits/e0a855ff8ab75db4bd34290274aea25dbe7641c5#diff-4191f5e37494ca6155f14396af0beb3c85994d231580ee4d58ccd3b691fe5696R348) in `TestParse` and make `removeLibraryDuplicates` function not exportable.

knqyf263: Oh, yes, I was confused with UniqBy. How about that? https://github.com/samber/lo#uniqby

DmitriyLewen: I need to check docs better... It works! Thanks for the advice!
Updated in https://github.com/aquasecurity/go-dep-parser/pull/221/commits/9346bf2ab153f40c80051084bee72f63eb8416a0

knqyf263 (approved):

PR #222, opened 2023-06-05
fix: change the type of the devDependencies field

nikpivkin:
I changed the type of the `devDependencies` field, because in v6 lock file it is a structure and not a string.

Close [#4559](https://github.com/aquasecurity/trivy/issues/4559)


knqyf263:
Hi @nikpivkin, thanks for the PR. We usually close the relevant issue when we merge PR in Trivy. `Close xxx` automatically closes the issue. Would you change it to "Issue https://github.com/aquasecurity/trivy/issues/4559"?

nikpivkin:
Hi @knqyf263. Thanks for the clarification, fixed.

knqyf263 (approved):
LGTM, thanks!

knqyf263:
@nikpivkin Can you please open a PR in Trivy? It just bumps the go-dep-parser version. Please note that the PR title and description should describe what is actually added or fixed. It should not be like "bump go-dep-parser".

nikpivkin:
@knqyf263 sure

PR #225, opened 2023-06-09
feat(java): capture licenses from pom.xml

nikpivkin:
## Description
At the moment, the licenses specified in pom.xml , do not parse
## Related Issues
- https://github.com/aquasecurity/trivy/issues/3812

### DIFF
--- a/pkg/java/pom/parse_test.go
+++ b/pkg/java/pom/parse_test.go
@@ -32,7 +32,7 @@ func TestPom_Parse(t *testing.T) {
 				{
 					Name:    "com.example:happy",
 					Version: "1.0.0",
-					License: "Apache 2.0",
+					License: "BSD-3-Clause",
 				},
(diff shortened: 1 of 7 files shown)


pkg/java/pom/parse_test.go
```diff
@@ -32,6 +32,7 @@ func TestPom_Parse(t *testing.T) {
 				{
 					Name:    "com.example:happy",
 					Version: "1.0.0",
+					License: "Apache 2.0",
```
knqyf263, line 35: Looks like all licenses are `Apache 2.0` other than multiply licenses. Can we change one of them to another license so that we can make sure the test identifies licenses correctly? This prevents bugs where values are not updated properly, and all values are the same by mistake.

nikpivkin: done

PR #231, opened 2023-06-14
fix(nodejs): parse package.json without name or version

nikpivkin:
Since the name and version are [optional](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name), we should consider package.json valid without these fields.

See https://github.com/aquasecurity/trivy/issues/4629


knqyf263:
Not ready for review?

nikpivkin:
@knqyf263 not yet

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,16 +39,19 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
...
-	// https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name
-	if pkgJSON.Name == "" || pkgJSON.Version == "" {
+	if pkgJSON.Name == "" && pkgJSON.Version == "" && !pkgJSON.hasContent() {
 		return Package{}, nil
 	}
 
+	name := pkgJSON.Name
+	if name == "" {
+		name = "no-package-name"
+	}
```
afdesk, line 49: also, let's add a comment here

afdesk (changes requested):

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,16 +57,19 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
 		return Package{}, xerrors.Errorf("JSON decode error: %w", err)
 	}
 
-	// Name and version fields are optional
-	// https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name
```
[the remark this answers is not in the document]
nikpivkin: done

pkg/nodejs/packagejson/parse.go
```diff
@@ -17,16 +19,36 @@ type packageJSON struct {
 	OptionalDependencies map[string]string `json:"optionalDependencies"`
 }
 
+func (p packageJSON) hasContent() bool {
+	return parseLicense(p.License) != "" || p.Dependencies != nil || p.OptionalDependencies != nil
```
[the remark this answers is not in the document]
nikpivkin: done

pkg/nodejs/packagejson/parse.go
nikpivkin: done.

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,16 +39,22 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
...
+	if pkgJSON.Name == "" && pkgJSON.Version == "" && !pkgJSON.hasContent() {
+		return Package{}, nil
+	}
+
+	name := pkgJSON.Name
 	// Name and version fields are optional
 	// https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name
-	if pkgJSON.Name == "" || pkgJSON.Version == "" {
-		return Package{}, nil
+	// if the name is missing, a dummy constant is used to identify this package.
```
afdesk, line 49:
```suggestion
// In cases where the name is not provided, but there is valuable content present, 
// we utilize a placeholder constant to identify this package.
```

afdesk (changes requested):

afdesk:
LGTM now.

pkg/nodejs/packagejson/parse.go
nikpivkin: done

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,16 +35,26 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
...
+
+	name := pkgJSON.Name
 	// Name and version fields are optional
 	// https://docs.npmjs.com/cli/v9/configuring-npm/package-json#name
-	if pkgJSON.Name == "" || pkgJSON.Version == "" {
-		return Package{}, nil
+	// In cases where the name is not provided, but there is valuable content present,
+	// we utilize a placeholder constant to identify this package.
+	if name == "" {
+		name = "no-package-name"
```
DmitriyLewen, line 51: Perhaps we need to use constant for `no-package-name` and compare package name with this constant in Trivy.
I am not sure that we need to put package without name to Trivy result.
If i understand correctly - that is why we skipped those `package.json` files previous.

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,16 +35,26 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
 		return Package{}, xerrors.Errorf("JSON decode error: %w", err)
 	}
 
+	license := parseLicense(pkgJSON.License)
+
+	hasContent := license != "" || len(pkgJSON.Dependencies) > 0 || len(pkgJSON.OptionalDependencies) > 0
```
DmitriyLewen, line 40: it looks as separate function.
In the future we can add more options here(e.g. `workspaces`).

DmitriyLewen (commented):
Hello @nikpivkin 
I left some comments.
Can you take a look?

pkg/nodejs/packagejson/parse.go
afdesk: the shared constant looks good.

afdesk: about the second question - we're discussing with @knqyf263

knqyf263: I'm confused. Why do we need to set `no-package-name`? What if just skipping empty IDs?

knqyf263: I mean we can skip it in the package.json analyzer. We should not skip it in the yarn analyzer.

afdesk: there are `package.json` without `name` or/and `version`, because these fields are optional. 
but such files can contain data about licenses and dependencies.
usually it's not a package for publishing, only for development. this case: https://github.com/aquasecurity/trivy/issues/4629

Trivy skips packages without ID: https://github.com/aquasecurity/trivy/blob/0e01851e9e5f18bff5e8306ead569ca1617cc9d4/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go#L34
so if there is useful content, we'll add a placeholder.

knqyf263: Should not skip here
https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/analyzer/language/nodejs/yarn/yarn.go#L172

Should skip here (already skipped)
https://github.com/aquasecurity/trivy/blob/main/pkg/fanal/analyzer/language/nodejs/pkg/pkg.go#L33-L36

afdesk: yes, you're right. package.json analyzer is not relevant for yarn.lock.

DmitriyLewen (approved):

pkg/nodejs/packagejson/parse.go
afdesk: I misunderstood the situation. I was worried about losing valuable data when parsing package.json, but we actually skip most of the fields. Having a License without a name or version doesn't make sense.

pkg/nodejs/packagejson/parse.go
```diff
@@ -35,18 +42,24 @@ func (p *Parser) Parse(r io.Reader) (Package, error) {
 		return Package{}, xerrors.Errorf("JSON decode error: %w", err)
 	}
 
+	license := parseLicense(pkgJSON.License)
+	if !pkgJSON.hasContent(license) {
+		return Package{}, nil
+	}
```
knqyf263, line 48: As @afdesk said, I don't think we need this logic. Returning a license only seems to be useless.
https://github.com/aquasecurity/go-dep-parser/pull/231#discussion_r1234800190

knqyf263 (approved):

PR #232, opened 2023-06-15
feat(nodejs): add the workspaces field to the package

nikpivkin:
Trivy needs to know about workspaces to analyze them.
See https://github.com/aquasecurity/trivy/issues/4633


DmitriyLewen (changes requested):
Hello @nikpivkin 
Thanks for your work!

Can you add `workspaces` field to test file and update testcase? 
This may be needed later.

DmitriyLewen (approved):
Thanks! Approved.

PR #235, opened 2023-06-20
fix(pom): don't skip parent when version is property

DmitriyLewen:
## Description
There is case when parent version is property (e.g. `${revision}`). We need to try find of this parent by relative path.

More information [here](https://github.com/aquasecurity/trivy/discussions/4670)

### DIFF
--- a/pkg/java/pom/pom.go
+++ b/pkg/java/pom/pom.go
@@ -26,7 +26,7 @@ func (p *pom) inherit(result analysisResult) {
 	p.content.GroupId = art.GroupID
 	p.content.ArtifactId = art.ArtifactID
 
-	if versionIsProperty(art.Version.String()) {
+	if isProperty(art.Version.String()) {
 		p.content.Version = evaluateVariable(art.Version.String(), p.content.Properties, nil)
 	} else {
 		p.content.Version = art.Version.String()
--- a/pkg/java/pom/pom.go
+++ b/pkg/java/pom/pom.go
@@ -25,7 +25,12 @@ func (p *pom) inherit(result analysisResult) {
 
 	p.content.GroupId = art.GroupID
 	p.content.ArtifactId = art.ArtifactID
-	p.content.Version = art.Version.String()
+
+	if versionIsProperty(art.Version.String()) {
+		p.content.Version = evaluateVariable(art.Version.String(), p.content.Properties, nil)
+	} else {
+		p.content.Version = art.Version.String()
+	}
 }
 
 func (p pom) properties() properties {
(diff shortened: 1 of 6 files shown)


pkg/java/pom/utils.go
```diff
@@ -9,3 +12,10 @@ func isDirectory(path string) (bool, error) {
 	}
 	return fileInfo.IsDir(), err
 }
+
+func versionIsProperty(version string) bool {
```
knqyf263, line 16: We may want to use this function in other places than values.

```suggestion
func isProperty(s string) bool {
```

DmitriyLewen: It make sense. Changed in https://github.com/aquasecurity/go-dep-parser/pull/235/commits/cf34368af151f49d98ae63e6a94da3a3b72c89f6.

pkg/java/pom/pom.go
```diff
@@ -25,7 +25,12 @@ func (p *pom) inherit(result analysisResult) {
 
 	p.content.GroupId = art.GroupID
 	p.content.ArtifactId = art.ArtifactID
```
knqyf263, line 27: Can't artifactId be a property?

DmitriyLewen: I have not seen information or examples of this.
Property in `Version` is needed to change version for all nested pom files.
For `ArtifactID` this doesn't make much sense, because you don't need to change `ArtifactID` from time to time. You create `GroupID` and `ArtifactID` once and use them.

DmitriyLewen: https://maven.apache.org/maven-ci-friendly.html
`Starting with Maven 3.5.0-beta-1 you can use the ${revision}, ${sha1} and/or ${changelist} as placeholders for the version in your pom file.`

DmitriyLewen: hm.. it is strange. This works, but maven returns a warning about this case:
pom.xml:
```
...
    <artifactId>${revision}-test-revision</artifactId>    
...
```
maven:
```
➜  ali-dbhub-server-domain-support mvn dependency:tree | grep test-revision
[WARNING] Some problems were encountered while building the effective model for com.alibaba.dbhub:1.0.0-SNAPSHOT-test-revision:jar:1.0.0-SNAPSHOT
[WARNING] 'artifactId' contains an expression but should be a constant. @ com.alibaba.dbhub:${revision}-test-revision:${revision}, /home/dmitriy/work/temp/4670/Chat2DB-bd1edda4d485fab2e06632fb50716748353153f0/ali-dbhub-server/ali-dbhub-server-domain/ali-dbhub-server-domain-support/pom.xml, line 12, column 17
...
[INFO] com.alibaba.dbhub:1.0.0-SNAPSHOT-test-revision:jar:1.0.0-SNAPSHOT
```

knqyf263: Thanks for the prompt testing! If maven warns it, we don't have to support it. I'll approve it.

knqyf263 (approved):

PR #239, opened 2023-06-30
fix(java): inherit licenses from parents

nikpivkin:
We have to inherit the licenses of the parent pom.xml. https://maven.apache.org/pom.html#Inheritance

### DIFF
--- a/pkg/java/pom/artifact.go
+++ b/pkg/java/pom/artifact.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/aquasecurity/go-dep-parser/pkg/log"
+	"github.com/samber/lo"
 	"golang.org/x/exp/slices"
 )
 
@@ -18,7 +19,7 @@ type artifact struct {
 	GroupID    string
 	ArtifactID string
 	Version    version
-	License    string
+	Licenses   []string
 
 	Exclusions map[string]struct{}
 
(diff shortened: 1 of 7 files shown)


pkg/java/pom/artifact.go
```diff
@@ -43,12 +44,26 @@ func (a artifact) Equal(o artifact) bool {
...
+	})}
+}
+
 func (a artifact) Inherit(parent artifact) artifact {
 	// inherited from a parent
 	if a.GroupID == "" {
 		a.GroupID = parent.GroupID
 	}
 
+	if len(a.Licenses) == 0 {
```
DmitriyLewen, line 63: I am not sure about this.
Do we need to get licenses from each parents?

Example:
We have `base pom` (contains `Apatch-1.0` license) with `package1`(uses `Apatch-1.0`) and `package2`(uses `Apatch-1.0`). `parent` also contains `Apatch1.0` (`package2` in `dependencyManagement`).
image

After a while, `another-group` (parent pom) updates version of `package2`(uses `Apatch-2.0`) and updates `licenses` field in parent pom.
Owner of `base pom` just updated parent version:
image

In this case i see 2 options:
1) This is mistake of `bases pom` owner - he should update licenses when updating parent version.
2) We need to inherit license from parent and include 2 licenses in result

@nikpivkin @knqyf263 wdyt?

nikpivkin: @DmitriyLewen the license field is overwritten in the child pom.xml. I was guided by the output of the [help:effective-pom](https://maven.apache.org/plugins/maven-help-plugin/effective-pom-mojo.html) command.

knqyf263: I didn't know `help:effective-pom`. It looks really useful.

DmitriyLewen (approved):
LGTM

pkg/java/pom/artifact.go
DmitriyLewen: okay, thanks for information. Let's stay this solution.

knqyf263 (approved):

PR #244, opened 2023-08-02
fix(npm): add init of root deps

DmitriyLewen:
## Description
There is case when root `Dependencies` field (packages[""].Dependencies) is `nil`.
We need initialize this map to insert direct dependencies from links.

## Related Issues
- https://github.com/aquasecurity/trivy/issues/4923

### DIFF
--- a/pkg/nodejs/npm/parse.go
+++ b/pkg/nodejs/npm/parse.go
@@ -178,6 +178,13 @@ func resolveLinks(packages map[string]Package) {
 		return
 	}
 
+	// Init root Dependencies map
+	if packages[""].Dependencies == nil {
+		rootPackage := packages[""]
+		rootPackage.Dependencies = make(map[string]string)
+		packages[""] = rootPackage
+	}
+
 	workspaces := packages[""].Workspaces
 	for pkgPath, pkg := range packages {
 		for linkPath, link := range links {
--- a/pkg/nodejs/npm/parse.go
+++ b/pkg/nodejs/npm/parse.go
@@ -178,14 +178,12 @@ func resolveLinks(packages map[string]Package) {
 		return
 	}
 
-	// Init root Dependencies map
-	if packages[""].Dependencies == nil {
-		rootPackage := packages[""]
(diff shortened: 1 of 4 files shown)


pkg/nodejs/npm/parse.go
```diff
@@ -178,6 +178,13 @@ func resolveLinks(packages map[string]Package) {
 		return
 	}
 
+	// Init root Dependencies map
+	if packages[""].Dependencies == nil {
+		rootPackage := packages[""]
+		rootPackage.Dependencies = make(map[string]string)
+		packages[""] = rootPackage
```
knqyf263, line 185: Doesn't this work?

```suggestion
		packages[""].Dependencies = make(map[string]string)
```

DmitriyLewen: I wanted do this, but `packages[""]` is not "addressable".
https://stackoverflow.com/a/69006398

knqyf263: Interesting. It works, though. I guess that is because the map is pointer.
https://github.com/aquasecurity/go-dep-parser/blob/ca96da88ad7bc414138750b3b34955e83221480b/pkg/nodejs/npm/parse.go#L207

So, we probably want to define rootPkgs, update it, and overwrite `packages[""]` at the end.

```
rootPkgs := packages[""]
if rootPkgs == nil {
    rootPkgs.Dependencies = make(map[string]string)

...

workspaces := rootPkgs.Workspaces

...

			if isWorkspace(pkgPath, workspaces) {
				rootPkgs.Dependencies[pkgNameFromPath(linkPath)] = pkg.Version
			}

...

packages[""] = rootPkgs
```

DmitriyLewen: Done in https://github.com/aquasecurity/go-dep-parser/pull/244/commits/5db81fdc171aacb9a99d9b3e200b4502a6735333

I didn't add init for Workspaces, because we only use Workspaces in loop. This panic doesn't affect loops.

knqyf263 (approved):

PR #245, opened 2023-08-02
feat(swift): add Package.resolved file support

DmitriyLewen:
## Description
Add `Package.resolved` v1 and v2 file support.


pkg/swift/swift/testdata/happy-Package.resolved
```diff
@@ -0,0 +1,34 @@
+{
```
knqyf263, line 1: I want to write down how this file was generated somewhere.

knqyf263 (approved):

PR #250, opened 2023-08-23
fix(swift): add ID field

DmitriyLewen:
## Description
Add `ID` field for swift `Package.resolved` files.


knqyf263 (approved):

knqyf263:
Note that we add the `v` prefix for Go, like `github.com/aquasecurity/trivy@v0.44.0` since Go uses the schema. I'm not sure we want to comply with that for Swift as well. I think the current schema (without `v`) is fine, and we can change it anytime.

DmitriyLewen:
Agree with you!

PR #252, opened 2023-08-24
refactor(cocoapods): use `utils.PackageID` for ID

DmitriyLewen:
## Description
`Purl` uses `@` to separate package name and version - https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#cocoapods
Use same logic.

### DIFF
--- a/pkg/swift/cocoapods/parse.go
+++ b/pkg/swift/cocoapods/parse.go
@@ -1,7 +1,7 @@
 package cocoapods
 
 import (
-	"fmt"
+	"sort"
 	"strings"
 
 	"golang.org/x/exp/maps"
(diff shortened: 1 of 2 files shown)


knqyf263:
It is not considered important to have the same schema as PURL. Rather, it was my view that it would be better to have the same format as the package manager. Is there any advantage to this change?

DmitriyLewen:
`Cocoapods` doesn't have `ID` format:
 - `Podfile` uses `pod '<pkgName>', '<version>'` format.
 - `Podfile.lock` uses `<pkgName> (<version>)`.
Also there is no command to install pod (you need to add line to `Podfile`).
`pod list` command uses `<pkgName> <version>` format (e.g.`3DTouchHelper 0.1`).

I thought it would be better to use default format.
But now i understand that some users can use current format and maybe we don't need to change format without need.

knqyf263 (approved):
If CocoaPods doesn't have the ID format, it is better to use the PURL schema.

PR #256, opened 2023-09-19
fix(python): add `non-separable` prefix for python packaging

DmitriyLewen:
## Description
- `License` field can contain information about various licenses, license exceptions , etc. - https://packaging.python.org/en/latest/specifications/core-metadata/#license.
We need to mark these licenses so we don't have to split them in Trivy, because it is not possible.

- License Classifiers and License File fields are intended for multiple use. Add support for these cases.

### DIFF
--- a/pkg/python/packaging/parse.go
+++ b/pkg/python/packaging/parse.go
@@ -33,18 +33,31 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
 	if l := h.Get("License-Expression"); l != "" {
 		license = l
 	} else if l := h.Get("License"); l != "" {
-		license = l
+		// The license field can contain information about different licenses, license exceptions , etc.:
+		// https://packaging.python.org/en/latest/specifications/core-metadata/#license,
+		// but it is impossible to define a delimiter to separate them.
+		// Mark them, so we don't have to separate them later.
+		license = types.NonSeparableLicensePrefix + l
 	} else {
+		var licenses []string
+		// license classifiers are deprecated:
+		// https://peps.python.org/pep-0639/#deprecate-license-classifiers
 		for _, classifier := range h.Values("Classifier") {
 			if strings.HasPrefix(classifier, "License :: ") {
 				values := strings.Split(classifier, " :: ")
-				license = values[len(values)-1]
-				break
+				// there can be several classifiers with licenses
+				licenses = append(licenses, values[len(values)-1])
 			}
 		}
+		license = strings.Join(licenses, ", ")
 	}
 	if license == "" && h.Get("License-File") != "" {
-		license = "file://" + h.Get("License-File")
+		var licenseFiles []string
+		for _, licenseFile := range h.Values("License-File") {
+			// there can be several license files
+			licenseFiles = append(licenseFiles, "file://"+licenseFile)
+		}
+		license = strings.Join(licenseFiles, ", ")
 	}
 
 	return []types.Library{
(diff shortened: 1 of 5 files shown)


pkg/types/types.go
```diff
@@ -74,4 +74,6 @@ const (
 	RefVCS          RefType = "vcs"
 	RefIssueTracker RefType = "issue-tracker"
 	RefOther        RefType = "other"
+
+	NonSeparableLicensePrefix = "non-separable: "
```
DmitriyLewen, line 78: We already have [file://](https://github.com/aquasecurity/go-dep-parser/blob/6eff4752bcbe21782f5521ad234b0ea501ad8410/pkg/python/packaging/parse.go#L47) prefix.

Perhaps we want to create new struct for [License](https://github.com/aquasecurity/go-dep-parser/blob/6eff4752bcbe21782f5521ad234b0ea501ad8410/pkg/types/types.go#L13) field.

just as idea:
``` go
type License struct {
	Value string
	Type  licenseType
}

type licenseType string

const (
	fileLicenseType             licenseType = "file" // filename for license file
	textLicenseType             licenseType = "text" // text of license (for split, classifier, etc.)
	nonSeparableTextLicenseType licenseType = "nonSeparable-text" // text of license without possible to split
	formattedLicenseType        licenseType = "formattedLicenses" // formatted licenses (e.g. in SPDX format)
)
```

pkg/python/packaging/parse.go
```diff
@@ -33,18 +33,31 @@ func (*Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, e
 	if l := h.Get("License-Expression"); l != "" {
 		license = l
 	} else if l := h.Get("License"); l != "" {
-		license = l
+		// The license field can contain information about different licenses, license exceptions , etc.:
+		// https://packaging.python.org/en/latest/specifications/core-metadata/#license,
+		// but it is impossible to define a delimiter to separate them.
+		// Mark them, so we don't have to separate them later.
+		license = types.NonSeparableLicensePrefix + l
 	} else {
```
DmitriyLewen, line 41: I was thinking about checking `license classifiers` before `license`, but in this case we might miss some nuances (like license exclusion).
So I left the old order.

lior-orca:
@DmitriyLewen is this going to be picked up again soon? :)
Thanks in advance!

DmitriyLewen:
Hello @lior-orca
We need to wait for @knqyf263's decision.

@knqyf263 Can you take a look at these changes?

knqyf263:
Please give me some more time. I have some prioritized tasks.

lior-orca:
@knqyf263  any updated? :) 
cc @DmitriyLewen

pkg/types/types.go
knqyf263: It sounds reasonable. We can go with that approach.

lior-orca:
@DmitriyLewen any updates on this one?

DmitriyLewen:
Hello @lior-orca 
We decided to add a new structure for licenses - https://github.com/aquasecurity/go-dep-parser/pull/256#discussion_r1329663767.
So I'm working on that now.

pkg/types/types.go
DmitriyLewen: @knqyf263 I have added `licenseType` to this PR.
I also updated https://github.com/aquasecurity/trivy/pull/5211 to show the required changes for Trivy.
#5211 requires some refactoring. I will do this if you confirm that we are ready for these changes.

Issue #257, opened 2023-09-20
Incorrect .NET deps parsing


noqcks:
I originally posted this on the trivy repo, but figured it should actually be posted here. Original post: https://github.com/aquasecurity/trivy/discussions/5208

----
### Description

Trivy doesn't generate a correct .NET dependency tree in CycloneDX. Please see this gist for the reference .deps.json file im using https://gist.github.com/noqcks/49089249820126cbaabe59b70ba12ae4

See the desired and actual behaviour section

### Desired Behavior

Dependencies are listed for this package
```
{
  "ref": "pkg:nuget/Microsoft.Extensions.Options@2.2.0",
  "dependsOn": [
    "pkg:nuget/Microsoft.Extensions.DependencyInjection.Abstractions@2.2.0",
    "pkg:nuget/Microsoft.Extensions.Primitives@2.2.0",
    "pkg:nuget/System.ComponentModel.Annotations@4.5.0"
  ]
}
```

### Actual Behavior

The dependencies are empty.
```
{
  "ref": "pkg:nuget/Microsoft.Extensions.Options@2.2.0",
  "dependsOn": []
},
```

### Reproduction Steps

```bash
Copy the .deps.json file from here https://gist.github.com/noqcks/49089249820126cbaabe59b70ba12ae4

Run

trivy fs MyWebApp.deps.json --format cyclonedx
```
```


### Target

Filesystem

### Scanner

None

### Output Format

CycloneDX

### Mode

Standalone

### Debug Output

```bash
``
trivy fs MyWebApp.deps.json --format cyclonedx --debug
2023-09-18T09:13:42.744-0700	DEBUG	["cyclonedx" "spdx" "spdx-json" "github"] automatically enables '--list-all-pkgs'.
2023-09-18T09:13:42.745-0700	DEBUG	Severities: ["UNKNOWN" "LOW" "MEDIUM" "HIGH" "CRITICAL"]
2023-09-18T09:13:42.745-0700	DEBUG	Ignore statuses	{"statuses": null}
2023-09-18T09:13:42.746-0700	INFO	"--format cyclonedx" disables security scanning. Specify "--scanners vuln" explicitly if you want to include vulnerabilities in the CycloneDX report.
2023-09-18T09:13:42.759-0700	DEBUG	cache dir:  /Users/noqcks/Library/Caches/trivy
[...]

PR #258, opened 2023-09-20
Add dotnet deps parsing

noqcks:
Fixes: https://github.com/aquasecurity/go-dep-parser/issues/257

### DIFF
--- a/pkg/dotnet/core_deps/parse.go
+++ b/pkg/dotnet/core_deps/parse.go
@@ -1,14 +1,17 @@
 package core_deps
 
 import (
-	"github.com/liamg/jfather"
+	"fmt"
 	"io"
 	"strings"
+
+	"github.com/liamg/jfather"
 
 	"golang.org/x/xerrors"
 
 	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
 	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
 
 	"github.com/aquasecurity/go-dep-parser/pkg/log"
 )
@@ -17,6 +20,22 @@ type Parser struct{}
 
 func NewParser() types.Parser {
 	return &Parser{}
+}
+
+func splitNameVer(nameVer string) (name, version string) {
+	split := strings.Split(nameVer, "/")
+	if len(split) != 2 {
+		// Invalid name
+		log.Logger.Warnf("Cannot parse .NET library version from: %s", nameVer)
+		return "", ""
+	}
+	name = split[0]
+	version = split[1]
+	return
+}
+
+func joinNameVer(name, version string) (nameVer string) {
+	return fmt.Sprintf("%s/%s", name, version)
 }
 
 func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	}
 
 	var libraries []types.Library
-	for nameVer, lib := range depsFile.Libraries {
-		if !strings.EqualFold(lib.Type, "package") {
+	var deps []types.Dependency
+	for pkgNameVersion, target := range depsFile.Targets[depsFile.RuntimeTarget.Name] {
+		library, ok := depsFile.Libraries[pkgNameVersion]
+		if !ok {
 			continue
 		}
 
-		split := strings.Split(nameVer, "/")
-		if len(split) != 2 {
-			// Invalid name
-			log.Logger.Warnf("Cannot parse .NET library version from: %s", nameVer)
+		name, version := splitNameVer(pkgNameVersion)
+		if name == "" || version == "" {
 			continue
 		}
 
-		libraries = append(libraries, types.Library{
-			Name:      split[0],
-			Version:   split[1],
-			Locations: []types.Location{{StartLine: lib.StartLine, EndLine: lib.EndLine}},
+		lib := types.Library{
+			ID:        utils.PackageID(name, version),
+			Name:      name,
+			Version:   version,
+			Locations: []types.Location{{StartLine: library.StartLine, EndLine: library.EndLine}},
+		}
+
+		var childDeps []string
+		for depName, depVersion := range target.Dependencies {
+			nameVer := joinNameVer(depName, depVersion)
+			_, ok := depsFile.Libraries[nameVer]
+			if !ok {
+				continue
+			}
+			depID := utils.PackageID(depName, depVersion)
+			childDeps = append(childDeps, depID)
+		}
+
+		deps = append(deps, types.Dependency{
+			ID:        lib.ID,
+			DependsOn: childDeps,
 		})
+
+		libraries = append(libraries, lib)
 	}
 
-	return libraries, nil, nil
+	return libraries, deps, nil
 }
 
 type dotNetDependencies struct {
-	Libraries map[string]dotNetLibrary `json:"libraries"`
+	RuntimeTarget dotNetRuntimeTarget                `json:"runtimeTarget"`
+	Libraries     map[string]dotNetLibrary           `json:"libraries"`
+	Targets       map[string]map[string]dotNetTarget `json:"targets"`
+}
+
+type dotNetRuntimeTarget struct {
+	Name string `json:"name"`
+}
+
+type dotNetTarget struct {
+	Dependencies map[string]string   `json:"dependencies"`
+	Runtime      map[string]struct{} `json:"runtime"`
 }
 
 type dotNetLibrary struct {
--- a/pkg/dotnet/core_deps/parse_test.go
+++ b/pkg/dotnet/core_deps/parse_test.go
@@ -15,19 +15,251 @@ import (
 
 func TestParse(t *testing.T) {
 	vectors := []struct {
-		file    string // Test input file
-		want    []types.Library
-		wantErr string
+		file     string // Test input file
+		wantLibs []types.Library
+		wantDeps []types.Dependency
+		wantErr  string
 	}{
 		{
-			file: "testdata/ExampleApp1.deps.json",
-			want: []types.Library{
-				{Name: "Newtonsoft.Json", Version: "13.0.1", Locations: []types.Location{{StartLine: 33, EndLine: 39}}},
+			file: "testdata/MyExample.deps.json",
+			wantLibs: []types.Library{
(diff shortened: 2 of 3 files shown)


DmitriyLewen:
Hello @noqcks 
Thanks for your work!

I will check your PR and write to you.

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+		if name == "" || version == "" {
 			continue
 		}
 
-		libraries = append(libraries, types.Library{
-			Name:      split[0],
-			Version:   split[1],
-			Locations: []types.Location{{StartLine: lib.StartLine, EndLine: lib.EndLine}},
+		lib := types.Library{
+			ID:        utils.PackageID(name, version),
```
DmitriyLewen, line 66: `.Net` uses `<pkg_name>/<pkg_version>` format.
Better to use this format - https://github.com/aquasecurity/go-dep-parser/pull/252#issuecomment-1691707417

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	}
 
 	var libraries []types.Library
-	for nameVer, lib := range depsFile.Libraries {
-		if !strings.EqualFold(lib.Type, "package") {
+	var deps []types.Dependency
+	for pkgNameVersion, target := range depsFile.Targets[depsFile.RuntimeTarget.Name] {
+		library, ok := depsFile.Libraries[pkgNameVersion]
```
DmitriyLewen, line 55: Can we discuss about this:

It looks like we can take package name and version from `pkgNameVersion`.
Do we need to check "Libraries"?
Is this necessary for backwards compatibility?

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
 
-		libraries = append(libraries, types.Library{
-			Name:      split[0],
-			Version:   split[1],
-			Locations: []types.Location{{StartLine: lib.StartLine, EndLine: lib.EndLine}},
+		lib := types.Library{
+			ID:        utils.PackageID(name, version),
+			Name:      name,
+			Version:   version,
+			Locations: []types.Location{{StartLine: library.StartLine, EndLine: library.EndLine}},
```
DmitriyLewen, line 69: `Libraries` array doesn't have info about child dependencies. 
I think we need to use `targets` location.

pkg/dotnet/core_deps/parse.go
```diff
@@ -19,6 +22,22 @@ func NewParser() types.Parser {
...
+func splitNameVer(nameVer string) (name, version string) {
+	split := strings.Split(nameVer, "/")
+	if len(split) != 2 {
+		// Invalid name
+		log.Logger.Warnf("Cannot parse .NET library version from: %s", nameVer)
+		return "", ""
+	}
+	name = split[0]
+	version = split[1]
+	return
```
DmitriyLewen, line 34: Let's use same format(previously you used `return "", ""`)
```suggestion
	return split[0], split[1]
```

pkg/dotnet/core_deps/parse.go
```diff
@@ -19,6 +22,22 @@ func NewParser() types.Parser {
...
+}
+
+func joinNameVer(name, version string) (nameVer string) {
```
DmitriyLewen, line 37: You can rename this function and use it to create ID.
```suggestion
func packageID(name, version string) string {
```

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+				continue
+			}
+			depID := utils.PackageID(depName, depVersion)
+			childDeps = append(childDeps, depID)
+		}
+
+		deps = append(deps, types.Dependency{
+			ID:        lib.ID,
+			DependsOn: childDeps,
 		})
```
DmitriyLewen, line 86: saving deps without child deps makes no sense.
```suggestion
		if len(childDeps) > 0{
			deps = append(deps, types.Dependency{
				ID:        lib.ID,
				DependsOn: childDeps,
			})
		}
```

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,30 +50,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	}
 
 	var libraries []types.Library
-	for nameVer, lib := range depsFile.Libraries {
-		if !strings.EqualFold(lib.Type, "package") {
+	var deps []types.Dependency
+	for pkgNameVersion, target := range depsFile.Targets[depsFile.RuntimeTarget.Name] {
```
DmitriyLewen, line 54: I don't have a lot of exp with .Net, but found next case:
`RuntimeTarget.Name` from depsFile != `RuntimeTarget.Name` from `targets`:
https://github.com/dotnet/cli/blob/f0ee1e44114f161ce9268103d18ce32d2a995d0f/Documentation/specs/runtime-configuration-file.md?plain=1#L48
https://github.com/dotnet/cli/blob/f0ee1e44114f161ce9268103d18ce32d2a995d0f/Documentation/specs/runtime-configuration-file.md?plain=1#L61

Is this possible? or is this a mistake?

pkg/dotnet/core_deps/parse_test.go
```diff
@@ -15,19 +15,251 @@ import (
...
-		want    []types.Library
-		wantErr string
+		file     string // Test input file
+		wantLibs []types.Library
+		wantDeps []types.Dependency
+		wantErr  string
 	}{
+		{
+			file: "testdata/MyExample.deps.json",
+			wantLibs: []types.Library{
```
DmitriyLewen, line 25: We try to make test cases smaller for better support.
Can you use a library with fewer dependencies and less depth?

pkg/dotnet/core_deps/parse_test.go
```diff
@@ -15,19 +15,251 @@ import (
...
+				{ID: "MyWebApp@1.0.0", Name: "MyWebApp", Version: "1.0.0", Locations: []types.Location{{StartLine: 148, EndLine: 152}}},
```
DmitriyLewen, line 26: Is there reason to add `project` to result?
We didn't include this earlier: https://github.com/aquasecurity/go-dep-parser/blob/6eff4752bcbe21782f5521ad234b0ea501ad8410/pkg/dotnet/core_deps/parse.go#L35

DmitriyLewen (commented):
Hello @noqcks 
Left some comments. Take a look, when you have time, please

pkg/dotnet/core_deps/parse.go
noqcks: I believe this is a mistake. I've never been able to find another example of this occurring in a real .json file (not an example in a `.md`)

pkg/dotnet/core_deps/parse.go
noqcks: I think I did this initially because I wanted to keep track of whether a Library was a project or not and then be able to set the CycloneDX `componentType` in response, but I don't think go-dep-parser keeps track of this? I don't see any field on the `Library` struct that would allow us to set something resembling `componentType`

noqcks: So, in this case, we don't need to check Libraries at all

pkg/dotnet/core_deps/parse_test.go
noqcks: yes, because otherwise you can't extract a full and complete dependency tree if you need to parse the SBOM

noqcks:
@DmitriyLewen ready for another review

pkg/dotnet/core_deps/parse.go
DmitriyLewen: > I don't think go-dep-parser keeps track of this

We set `componentType` in Trivy. 

>  I wanted to keep track of whether a Library was a project or not 

We currently take only packages - https://github.com/aquasecurity/go-dep-parser/blob/1237b47625e1c1e543dcc899e112c3b1e3d8b7ff/pkg/dotnet/core_deps/parse.go#L35
For project we use `application` type with filepath as name:
```
➜  trivy -d fs -f cyclonedx pkg/dotnet/core_deps/testdata/ExampleApp1.deps.json | jq '.components[] | select(.type == "application")'
...
{
  "bom-ref": "908ce44b-3644-44ce-b282-8f0d44566e20",
  "type": "application",
  "name": "ExampleApp1.deps.json",
  "properties": [
    {
      "name": "aquasecurity:trivy:Class",
      "value": "lang-pkgs"
    },
    {
      "name": "aquasecurity:trivy:Type",
      "value": "dotnet-core"
    }
  ]
}
```

pkg/dotnet/core_deps/parse.go
DmitriyLewen: Okey. Thanks!

pkg/dotnet/core_deps/parse_test.go
DmitriyLewen: This makes sense for SBOM, but can be confusing when used by default (like` json` format).
We save project in dependencies. This can be confusing.

Perhaps we need to add new type or field for `Library` for project name/version. But I think it is change for another PR.

We use `application` componentType for SBOM - https://github.com/aquasecurity/go-dep-parser/pull/258#discussion_r1353970059
Previously this was enough.

pkg/dotnet/core_deps/parse.go
```diff
@@ -31,40 +47,60 @@ func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+				ID:        lib.ID,
+				DependsOn: childDeps,
+			})
+		}
+
+		libraries = append(libraries, lib)
 	}
 
-	return libraries, nil, nil
+	return libraries, deps, nil
```
DmitriyLewen, line 83: `Libraries` and `Dependencies` can be sorted.
```suggestion
	sort.Sort(types.Libraries(libs))
	sort.Sort(types.Dependencies(deps))
	
	return libraries, deps, nil
```

After these changes you can remove sorting from tests.

DmitriyLewen (commented):
I left 2 small comment. Take a look, please.

Also i understood that `MyExample.deps.json` already contains dependencies.
Do we need new test file or will `MyExample.deps.json` be enough?

PR #261, opened 2023-10-05
fix(python-packaging): skip MIME error, if name and version found

DmitriyLewen:
## Description
`Golang` updated net/textproto package - https://cs.opensource.google/go/go/+/a6642e67e16b9d769a0c08e486ba08408064df19

After that we get error if MIME key or values contain bytes non-allowed by RFC 7230.

But if required fields (`name` and `version` in this case) found - we don't need to stop parsing.

Related Issues:
- aquasecurity/trivy/issues/5335

### DIFF
--- a/pkg/python/packaging/parse.go
+++ b/pkg/python/packaging/parse.go
@@ -2,6 +2,7 @@ package packaging
 
 import (
 	"bufio"
+	"github.com/aquasecurity/go-dep-parser/pkg/log"
 	"net/textproto"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/python/packaging/parse.go
```diff
@@ -23,8 +23,17 @@ func NewParser() types.Parser {
...
+	if err != nil {
+		if name == "" || version == "" {
+			// Some unnecessary headers for this case may contain bytes in the
+			// key or value outside the set allowed by RFC 7230
+			// In this case we get error:
+			// cf. https://cs.opensource.google/go/go/+/a6642e67e16b9d769a0c08e486ba08408064df19
+			// If `name` and `version` are found, we don't need to stop
+			return nil, nil, xerrors.Errorf("read MIME error: %w", err)
+		}
+		log.Logger.Debugf("Package 'name' and 'version' were found, but a MIME reading error occurs: %s", err)
```
DmitriyLewen, line 36: I am not sure - perhaps we need to use `Warn`.

knqyf263 (approved):

PR #267, opened 2023-10-26
fix(pom): compare ArtifactIDs for base and parent pom's when `relativePath` field is used

DmitriyLewen:
## Description
We need to compare `ArtifactID`s for base and parent pom's when `relativePath == ../pom.xml` is used.
it is needed to avoid cases when we get incorrect parent or get infinity loop (if `../pom.xml` is scanned pom.xml).

We can inherit `GroupID`. Also `Version` can contains property.
So we need to compare only `ArtifactID`.

New test contains simple example - https://github.com/DmitriyLewen/go-dep-parser/blob/7d2c865da8912f9f3ea475cc94da61c819dcae0d/pkg/java/pom/parse_test.go#L987-L999
Example from real life - aquasecurity/trivy/discussions/5445

## Related Issues
- aquasecurity/trivy/issues/5453

### DIFF
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -510,6 +510,13 @@ func (p *parser) tryRelativePath(parentArtifact artifact, currentPath, relativeP
 		return nil, err
 	}
 
+	// We can inherit GroupID from parent (requires p.analyze function to get GroupID)
+	// Version can contain a property (requires p.analyze function to get version)
+	// But ArtifactID must be the same for parent and base poms
+	// This check is necessary to avoid an infinite loop when using relatedPath or `../pom.xml`
+	if pom.artifact().ArtifactID != parentArtifact.ArtifactID {
+		return nil, xerrors.New("'parent.relativePath' points at wrong local POM")
+	}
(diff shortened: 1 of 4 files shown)


pkg/java/pom/parse.go
```diff
@@ -510,6 +510,13 @@ func (p *parser) tryRelativePath(parentArtifact artifact, currentPath, relativeP
 		return nil, err
 	}
 
+	// We can inherit GroupID from parent (requires p.analyze function to get GroupID)
+	// Version can contain a property (requires p.analyze function to get version)
+	// But ArtifactID must be the same for parent and base poms
```
knqyf263, line 515: What do you mean? The parent artifact id and the base artifact id won't match.
Do you mean, "The artifact id of the parent pom in the base pom and the actual artifact id in the parent pom must match."?

DmitriyLewen: Right!

But i see that this comment may be confusing.
I updated comment in https://github.com/aquasecurity/go-dep-parser/pull/267/commits/cedc6ceb39982a8117b1daa533f5f6c43c87e9be :
Take a look, please.

knqyf263 (approved):

PR #268, opened 2023-10-29
adding packages.props and Directory.packages.props parser

yuriShafet:


pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+
+type propsPackageVersionEntry struct {
+	XMLName         xml.Name `xml:"PackageVersion"`
+	propsPackageEntry
+}
+
+type propsItemGroup struct {
+	XMLName  xml.Name              `xml:"ItemGroup"`
+	ReferencePackages []propsPackageReferenceEntry `xml:"PackageReference"`
+	VersionPackages []propsPackageVersionEntry `xml:"PackageVersion"`
```
DmitriyLewen, line 34: Looks like old and new `*.packages.props` files only use `ReferencePackages`.
Can you add link to docs/example for `VersionPackages` and add this case in tests?

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+
+	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
+)
+
+	
+
+
+type propsPackageEntry struct{
```
DmitriyLewen, line 15: I think we can omit `props` and `propsPackage` for structs.
```suggestion
type entry struct{
```

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+
+type propsItemGroup struct {
+	XMLName  xml.Name              `xml:"ItemGroup"`
+	ReferencePackages []propsPackageReferenceEntry `xml:"PackageReference"`
+	VersionPackages []propsPackageVersionEntry `xml:"PackageVersion"`
+}
+
+type propsProject struct {
+	XMLName xml.Name `xml:"Project"`
+	ItemGroups []propsItemGroup `xml:"ItemGroup"`
```
DmitriyLewen, line 39: ItemGroups looks as single tag. 
If you have seen `ItemGroups` array in docs or some examples - can you add comment about this and add that in tests?

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type Parser struct{}
+
+func NewParser() types.Parser {
+	return &Parser{}
+}
+
+func (p *Parser) addPackage(libs* []types.Library, pkg propsPackageEntry) {
```
DmitriyLewen, line 48: You save libs not explicitly 
This is not good and can be confusing.

Better to explicitly return lib.
```suggestion
func parsePackage(pkg propsPackageEntry) types.Library {
```

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+			Name:    pkg.IncludePackageName,
+			Version: pkg.Version,
+		}
+	}else{
+		lib = types.Library{
+			Name:    pkg.UpdatePackageName,
+			Version: pkg.Version,
+		}
+	}
+	*libs = append(*libs, lib)
```
DmitriyLewen, line 65:
```suggestion
	name := pkg.IncludePackageName
	if pkg.UpdatePackageName != "" {
		name = pkg.UpdatePackageName
	}
	return types.Library{
	        ID:      utils.PackageID(name, pkg.Version),
		Name:    name,
		Version: pkg.Version,
	}
```

pkg/nuget/packages_props/parse.go
DmitriyLewen, line 34: looks like we can just use `propsPackageEntry`:
```suggestion
	ReferencePackages []propsPackageEntry `xml:"PackageReference"`
	VersionPackages   []propsPackageEntry `xml:"PackageVersion"`
```

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+		for _, refPkg := range itemGroup.ReferencePackages{
+			var pkg = propsPackageEntry{refPkg.Version, refPkg.UpdatePackageName, refPkg.IncludePackageName}
+			p.addPackage(&libs, pkg)
+		}
+
+		for _, verPkg := range itemGroup.VersionPackages{
+			var pkg = propsPackageEntry{verPkg.Version, verPkg.UpdatePackageName, verPkg.IncludePackageName}
+			p.addPackage(&libs, pkg)
+		}
+	}
```
DmitriyLewen, line 88: This way there will be no code duplication.
```suggestion
libs := make([]types.Library, 0)
	for _, itemGroup := range configData.ItemGroups {
		for _, pkg := range append(itemGroup.ReferencePackages, itemGroup.VersionPackages...) {
			lib := parsePackage(pkg)
			if lib.Name != "" {
				libs = append(libs, lib)
			}
		}
	}
```

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+	return utils.UniqueLibraries(libs), nil, nil
```
DmitriyLewen, line 89: Are there cases when one `*packages.props` file contains 2 same dependencies?
Perhaps this will be enough:
```suggestion
	sort.Sort(types.Libraries(libs))
	return libs, nil, nil
```

pkg/nuget/packages_props/parse_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+			got, _, err := config.NewParser().Parse(f)
+			if tt.wantErr != "" {
+				require.NotNil(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
+			}
+
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, tt.want, got)
```
DmitriyLewen, line 63: We use `Equal` function. let's try to match general style
```suggestion
			assert.Equal(t, tt.want, got)
```

pkg/nuget/packages_props/testdata/packages.props
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">  
+  <ItemGroup Label="Microsoft Nugets">
+    <PackageReference Update="Microsoft.Extensions.Configuration" Version="2.1.1" />
+    <PackageReference Update="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.1" />
+    <PackageReference Update="Microsoft.Extensions.Http" Version="3.2.1" />
+    <PackageReference WrongAttribute="Package1" Version="3.2.1" />
+    <WrongEntry Update="Package2" Version="3.2.1" />
```
DmitriyLewen, line 8: What is `WrongEntry`? 
Do we need to process this?

pkg/nuget/packages_props/parse.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+	}
+	*libs = append(*libs, lib)
+}
+
+
+func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var configData propsProject
+	if err := xml.NewDecoder(r).Decode(&configData); err != nil {
+		return nil, nil, xerrors.Errorf("failed to decode  .props file: %w", err)
```
DmitriyLewen, line 72:
```suggestion
		return nil, nil, xerrors.Errorf("failed to decode '*packages.props' file: %w", err)
```

DmitriyLewen (changes requested):
Hello @yuriShafet 
Thanks for work!

I left comments. Take a look, when you have time, please.

UPD:
1 more question - Can I use local packages? How does NuGet add them?

pkg/nuget/packages_props/parse.go
yuriShafet: Sure.
Packages props example:
https://github.com/dotnet/msbuild/blob/main/eng/Packages.props
Link from dotnet repo where these packages are parsed. Both Package Reference and Package Version are considered:
https://github.com/dotnet/roslyn-tools/blob/380ffede53738d15ed1384e0f52dd8a3d937d38b/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs#L495

pkg/nuget/packages_props/testdata/packages.props
yuriShafet: It is a test case, If Entry is not PackageRefernce or PackageVersion we should not consider it as a package.

pkg/nuget/packages_props/parse.go
yuriShafet: I think potentially it could contain same entries

pkg/nuget/packages_props/parse.go
yuriShafet: An example of Directory.packages.props using packageVersion
https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions

pkg/nuget/packages_props/parse.go
yuriShafet: As could be seen in implementation there could be several:
https://github.com/dotnet/roslyn-tools/blob/380ffede53738d15ed1384e0f52dd8a3d937d38b/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs#L494C41-L494C41
Ill add more tests

pkg/nuget/packages_props/parse.go
DmitriyLewen: Okay, thanks.
Can you add link to this example in comment, please.

Hm.. look like version can use variables, but we can't resolve them, right? https://github.com/dotnet/msbuild/blob/31c4d335325e858a3c4dba66d921d0e31bdee5ff/eng/Packages.props#L11
Maybe we need to remove them from result or consider finding these values from another file and merging result in Trivy (we use this logic to find licenses from nuspec files when parsing NuGet lock and config files - https://github.com/aquasecurity/trivy/blob/d0d956fdc1aa71ddf7f806888ea2b64d8441bfa1/pkg/fanal/analyzer/language/dotnet/nuget/nuget.go#L82)

Also i think it is better to use same name for fields from structure and xml tag
I mean renaming: ReferencePackages -> PackageReference and VersionPackages -> PackageVersion

pkg/nuget/packages_props/parse.go
DmitriyLewen: okay, thanks. Add this link to comment, please.

pkg/nuget/packages_props/parse.go
DmitriyLewen: Can you try to do that?
 I have guess that NuGet resolves these cases and doesn't allow these cases

pkg/nuget/packages_props/testdata/packages.props
DmitriyLewen: Got it! thanks.

pkg/nuget/packages_props/parse.go
yuriShafet: I could not find any documentation regarding that, but I tried adding a duplicate package  with same name and version in packages.props and the project compiled without any issues.

yuriShafet:
@DmitriyLewen I pushed updated version. Please resolve conversations that you think is resolved done and add more comments if necessary.

pkg/nuget/packages_props/parse.go
DmitriyLewen: Okay, thanks for this investigation!

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type entry struct {
+	Version            string `xml:"Version,attr"`
+	UpdatePackageName  string `xml:"Update,attr"`
+	IncludePackageName string `xml:"Include,attr"`
+}
+
+type propsItemGroup struct {
+	xml.Name              `xml:"ItemGroup"`
```
DmitriyLewen, line 21: We can remove this

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+		name = pkg.IncludePackageName
+	}
+	return types.Library{
+		ID:      utils.PackageID(name, pkg.Version),
+		Name:    name,
+		Version: pkg.Version,
+	}
+}
+
+func isNonEmptyNonVariableLib(lib types.Library) bool {
```
DmitriyLewen, line 49: This is a long name. What if we just use `skipLib`?

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	if isVariable(lib.Name) || isVariable(lib.Version) {
+		return false
+	}
+	return true
+
+}
+
+func isVariable(s string) bool {
+	trimmed := strings.Trim(s, " ")
```
DmitriyLewen, line 61: Do we need this Trim here?

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+		Name:    name,
+		Version: pkg.Version,
+	}
+}
+
+func isNonEmptyNonVariableLib(lib types.Library) bool {
+	if len(lib.Name) == 0 || len(lib.Version) == 0 {
+		return false
+	}
+	if isVariable(lib.Name) || isVariable(lib.Version) {
```
DmitriyLewen, line 53:
```suggestion
	// *packages.props files don't contain variable resolution information.
	// So we need to skip them.
	if isVariable(lib.Name) || isVariable(lib.Version) {
```

pkg/nuget/packages_props/parse.go
DmitriyLewen: Can you do the same for other structures?

pkg/nuget/packages_props/parse.go
DmitriyLewen: Sorry, I didn't think about that when I wrote this comment:
What if we just use `Library` name?
```
func library(pkg entry) types.Library {
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	// https://github.com/dotnet/roslyn-tools/blob/b4c5220f5dfc4278847b6d38eff91cc1188f8066/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs#L150
+	// Based on this documentation both legacy packages.props and Directory.packages.props are supported
+	// Directory.packages.props example: https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions
+
+	libs := make([]types.Library, 0)
+	for _, itemGroup := range configData.ItemGroups {
+		for _, refPkg := range append(itemGroup.PackageReferenceEntry, itemGroup.PackageVersionEntry...) {
+			var pkg = entry{refPkg.Version, refPkg.UpdatePackageName, refPkg.IncludePackageName}
+			var lib = parsePackage(pkg)
```
DmitriyLewen, line 78: you don't have to use `var`
```suggestion
			lib := parsePackage(pkg)
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, nil, xerrors.Errorf("failed to decode '*.packages.props' file: %w", err)
+	}
+	// https://github.com/dotnet/roslyn-tools/blob/b4c5220f5dfc4278847b6d38eff91cc1188f8066/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs#L150
+	// Based on this documentation both legacy packages.props and Directory.packages.props are supported
+	// Directory.packages.props example: https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions
+
+	libs := make([]types.Library, 0)
+	for _, itemGroup := range configData.ItemGroups {
+		for _, refPkg := range append(itemGroup.PackageReferenceEntry, itemGroup.PackageVersionEntry...) {
+			var pkg = entry{refPkg.Version, refPkg.UpdatePackageName, refPkg.IncludePackageName}
```
DmitriyLewen, line 77: `refPkg` is `entry`. You don't need to init new struct.
Just use `refPkg` in `parsePackage(refPkg)`

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,85 @@
...
+	"strings"
+
+	"golang.org/x/xerrors"
+
+	dio "github.com/aquasecurity/go-dep-parser/pkg/io"
+	"github.com/aquasecurity/go-dep-parser/pkg/types"
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
+)
+
+type entry struct {
```
DmitriyLewen, line 14: `PackageReference` and `PackageVersion` => `package`
```suggestion
type package struct {
```

DmitriyLewen (commented):
Hello @yuriShafet 
Left comments.

pkg/nuget/packagesprops/parse.go
yuriShafet: Yes. Just tested it. I can add spaces before and after package name and nuget restore works fine. Ill add more tests for that case

yuriShafet:
> Hello @yuriShafet Left comments.
> 
> Did you check local packages ([#268 (review)](https://github.com/aquasecurity/go-dep-parser/pull/268#pullrequestreview-1703440638))?

You mean packages that located in some local folder and nuget is configured to look into the folder or projects in same solution?
If you mean packages that located locally, it should work with that kind of packages.

pkg/nuget/packagesprops/parse.go
yuriShafet: package is reserved word, so I renamed it to pkg

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/aquasecurity/go-dep-parser/pkg/utils"
+)
+
+type pkg struct {
+	Version            string `xml:"Version,attr"`
+	UpdatePackageName  string `xml:"Update,attr"`
+	IncludePackageName string `xml:"Include,attr"`
+}
+
+type itemGroup struct {
```
DmitriyLewen, line 20:
```suggestion
// https://github.com/dotnet/roslyn-tools/blob/b4c5220f5dfc4278847b6d38eff91cc1188f8066/src/RoslynInsertionTool/RoslynInsertionTool/CoreXT.cs#L150
type itemGroup struct {
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,86 @@
...
+	if p.IncludePackageName != "" {
+		name = strings.Trim(p.IncludePackageName, " ")
+	}
+	version := strings.Trim(p.Version, " ")
+	return types.Library{
+		ID:      utils.PackageID(name, version),
+		Name:    name,
+		Version: version,
+	}
+}
```
DmitriyLewen, line 47:
```suggestion
func (p pkg) library() types.Library {
	// Update attribute is considered legacy, so preferring Include
	name := p.UpdatePackageName
	if p.IncludePackageName != "" {
		name = p.IncludePackageName
	}

	name = strings.TrimSpace(name)
	version := strings.TrimSpace(p.Version)
	return types.Library{
		ID:      utils.PackageID(name, version),
		Name:    name,
		Version: version,
	}
}
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,86 @@
...
+func shouldSkipLib(lib types.Library) bool {
+	if len(lib.Name) == 0 || len(lib.Version) == 0 {
+		return false
+	}
+	// *packages.props files don't contain variable resolution information.
+	// So we need to skip them.
+	if isVariable(lib.Name) || isVariable(lib.Version) {
+		return false
+	}
+	return true
```
DmitriyLewen, line 58: Inside this function you are using inverse logic.
```suggestion
func shouldSkipLib(lib types.Library) bool {
	if len(lib.Name) == 0 || len(lib.Version) == 0 {
		return true
	}
	// *packages.props files don't contain variable resolution information.
	// So we need to skip them.
	if isVariable(lib.Name) || isVariable(lib.Version) {
		return true
	}
	return false
}
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,86 @@
...
+
+}
+
+func isVariable(s string) bool {
+	trimmed := strings.Trim(s, " ")
```
DmitriyLewen, line 63: You trimmed name and version when created Library
```suggestion
```

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,86 @@
...
+
+	libs := make([]types.Library, 0)
+	for _, itemGroup := range configData.ItemGroups {
+		for _, pkg := range append(itemGroup.PackageReferenceEntry, itemGroup.PackageVersionEntry...) {
+			lib := library(pkg)
+			if shouldSkipLib(lib) {
+				libs = append(libs, lib)
+			}
+		}
+	}
```
DmitriyLewen, line 84: We can move comment to ItemGroups struct.
```suggestion
	var libs []types.Library
	for _, item := range configData.ItemGroups {
		for _, pkg := range append(item.PackageReferenceEntry, item.PackageVersionEntry...) {
			lib := pkg.library()
			if !shouldSkipLib(lib) {
				libs = append(libs, lib)
			}
		}
	}
```

DmitriyLewen (commented):
Unfortunately, GitHub doesn’t allow me to make changes to your PR (maybe it is because you are using `main` branch)

I left a few changes
Check that they don't break anything and merge them please

yuriShafet:
> Unfortunately, GitHub doesn’t allow me to make changes to your PR (maybe it is because you are using `main` branch)
> 
> I left a few changes Check that they don't break anything and merge them please

I merged your changes and ushed updated version. Thanks for the comments.

pkg/nuget/packagesprops/parse.go
```diff
@@ -0,0 +1,87 @@
...
+	return strings.HasPrefix(s, "$(") && strings.HasSuffix(s, ")")
+}
+
+func (p *Parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency, error) {
+	var configData project
+	if err := xml.NewDecoder(r).Decode(&configData); err != nil {
+		return nil, nil, xerrors.Errorf("failed to decode '*.packages.props' file: %w", err)
+	}
+
+	libs := make([]types.Library, 0)
```
DmitriyLewen, line 77: In this case you better to create `nil` value instead of empty types.Library slice ( there are cases when `nil` != `nil` - https://dev.to/arxeiss/in-go-nil-is-not-equal-to-nil-sometimes-jn8)
```suggestion
	var libs []types.Library
```

pkg/nuget/packagesprops/parse_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+			got, _, err := config.NewParser().Parse(f)
+			if tt.wantErr != "" {
+				require.NotNil(t, err)
+				assert.Contains(t, err.Error(), tt.wantErr)
+				return
+			}
+
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, tt.want, got)
```
DmitriyLewen, line 79: `utils.UniqueLibraries(libs)` also sorts libs. Better to use `assert.Equal` to see that libs has been sorted.
```suggestion
			assert.Equal(t, tt.want, got)
```

yuriShafet:
@DmitriyLewen 
Resolved the comments

DmitriyLewen (approved):
Great! Thanks!

PR looks good to me now.

Thanks for your work and sorry that i had to ask you to make all small changes.
To add options for maintainers to update your PR - it is better to create new PRs in a different branch (not the main).

@knqyf263 I approved this PR. Can you take a look, when you have time.

knqyf263:
@DmitriyLewen I trust you 👍

DmitriyLewen:
@yuriShafet Can you create PR for Trivy now?
Some help info - https://aquasecurity.github.io/trivy/v0.47/community/contribute/pr/

yuriShafet:
@DmitriyLewen Will Do. Thanks

PR #269, opened 2023-10-30
feat: create zap logger on demand

lebauce:
Importing one package causes a logger to be created and  Zap logger has a non negligible cost (a few hundred Kilobytes).

This PR simply defers the creation of the logger when it's actually used.

### DIFF
--- a/pkg/log/lazy.go
+++ b/pkg/log/lazy.go
@@ -1,0 +1,99 @@
+package log
+
+import "go.uber.org/zap"
+
+type lazyLogger struct {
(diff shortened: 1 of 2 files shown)


knqyf263:
Are you in a hurry? We're planning to replace zap with [slog](https://pkg.go.dev/log/slog) after Go 1.22 is out. This is because we have a policy supporting the two latest Go versions, and slog was added in Go 1.21.

lebauce:
@knqyf263 Thanks for the notice ! I can definitely wait

PR #276, opened 2023-11-22
feat(jar): add version for SearchByArtifactID function

DmitriyLewen:
## Description
Add version to `SearchByArtifactID` function (for [Client](https://github.com/aquasecurity/go-dep-parser/blob/fc7f2b470d8926796902748479355ac92e153aaa/pkg/java/jar/parse.go#L27-L31) interface) so that you can get the GroupId with only the required version.
See aquasecurity/trivy/issues/5627

### DIFF
--- a/pkg/java/jar/sonatype/sonatype.go
+++ b/pkg/java/jar/sonatype/sonatype.go
@@ -153,7 +153,7 @@ func (s Sonatype) SearchBySHA1(sha1 string) (jar.Properties, error) {
 	}, nil
 }
 
-func (s Sonatype) SearchByArtifactID(artifactID string) (string, error) {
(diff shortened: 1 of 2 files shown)


pkg/java/jar/sonatype/sonatype.go
```diff
@@ -153,7 +153,7 @@ func (s Sonatype) SearchBySHA1(sha1 string) (jar.Properties, error) {
 	}, nil
 }
 
-func (s Sonatype) SearchByArtifactID(artifactID string) (string, error) {
+func (s Sonatype) SearchByArtifactID(artifactID, _ string) (string, error) {
```
DmitriyLewen, line 156: In Trivy we use `trivy-java-db`.
Tell me if you want to update the sonatype logic as well.

knqyf263 (approved):

PR #280, opened 2023-12-11
fix: check Dev field for UniqueLibraries func

DmitriyLewen:
## Description
We need to check (and not overwrite) `Dev` field when retrieving uniq libs (`utils.UniqueLibraries` function).
See aquasecurity/trivy/issues/5532 for more details.


knqyf263 (approved):

knqyf263:
@DmitriyLewen Do we need to bump the go-dep-parser version in Trivy?

DmitriyLewen:
True, but we are blocked by https://github.com/aquasecurity/trivy/pull/5630
Because we updated `SearchByArtifactID` function in https://github.com/aquasecurity/go-dep-parser/pull/276

DmitriyLewen:
I created https://github.com/aquasecurity/trivy/pull/5837 to remember to update go-dep-parser in Trivy

PR #282, opened 2023-12-27
fix(pom): don't remove excluded deps from upper pom's

DmitriyLewen:
## Description
Fixed cases where we were removing excluded deps from upper pom's.
Example - https://github.com/aquasecurity/trivy/issues/5827
Test to better understand the fix - https://github.com/aquasecurity/go-dep-parser/blob/bf1b2d1cb5f5ad139c03d378674969ecd410d335/pkg/java/pom/parse_test.go#L689-L725

### DIFF
--- a/pkg/java/pom/pom.go
+++ b/pkg/java/pom/pom.go
@@ -4,6 +4,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"maps"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


pkg/java/pom/pom.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"maps"
```
DmitriyLewen, line 7: Wait #283

DmitriyLewen: #283 has been merged.

pkg/java/pom/pom.go
```diff
@@ -253,10 +254,12 @@ func (d pomDependency) Resolve(props map[string]string, depManagement, rootDepMa
...
-func (d pomDependency) ToArtifact(exclusions map[string]struct{}) artifact {
-	if exclusions == nil {
-		exclusions = map[string]struct{}{}
-	}
+func (d pomDependency) ToArtifact(ex map[string]struct{}) artifact {
+	// To avoid shadow adding exclusions to top pom's,
+	// we need to initialize a new map for each new artifact
+	// See `exclusions in child` test for more information
+	exclusions := map[string]struct{}{}
+	maps.Copy(exclusions, ex)
```
knqyf263, line 262: How about maps.Clone?

```suggestion
	exclusions := maps.Clone(ex)
```

DmitriyLewen: Looks like we will have enough of a hallow clone. Thank you!
Changed in https://github.com/aquasecurity/go-dep-parser/pull/282/commits/cdc2c7f5f7e7e092457c0c18762b35339495b7cf

knqyf263 (approved):

PR #285, opened 2024-01-18
fix(pom): recursive check all nested depManagements with `import` scope

DmitriyLewen:
## Description 
This PR corrects 2 related problems:
- we need to check all nested depManagements using `import` (recursively running `resolveDepManagement`).
- exclude shadow overwriting of props when some iterations of nested depManagements use the same props (e.g., ${project.version}).

See updated [import dependencyManagement test](https://github.com/aquasecurity/go-dep-parser/blob/b23678a961ed38e7f3dd0c08be7994733b427c1a/pkg/java/pom/parse_test.go#L553-L578) for more details.

## Related Issues
- https://github.com/aquasecurity/trivy/issues/5899


DmitriyLewen:
These 3 nested depManagements  and [dependency tag for 1 of them](https://github.com/aquasecurity/go-dep-parser/blob/6498684f82d379351b4a10e92e8029421a8d2d7f/pkg/java/pom/testdata/repository/org/example/example-dependency-management/3.3.3/example-dependency-management-3.3.3.pom#L27-L33) are required to show the need of [clone map properties](https://github.com/aquasecurity/go-dep-parser/blob/6498684f82d379351b4a10e92e8029421a8d2d7f/pkg/utils/utils.go#L61).

I found this problem after adding [recurcive](https://github.com/aquasecurity/go-dep-parser/blob/6498684f82d379351b4a10e92e8029421a8d2d7f/pkg/java/pom/parse.go#L414) - we overwrite the properties in the shadow.

knqyf263:
[Reverting your change](https://github.com/aquasecurity/go-dep-parser/pull/285/commits/226facd889b9e0f9ab1debe66ec11e74ad462c66) failed the test, meaning the simplified test can catch the edge case you found, right?

knqyf263:
BTW, I wanted to push the commits for testing to my fork, but pushed upstream by mistake. Sorry 😔

DmitriyLewen:
> [Reverting your change](https://github.com/aquasecurity/go-dep-parser/pull/285/commits/226facd889b9e0f9ab1debe66ec11e74ad462c66) failed the test, meaning the simplified test can catch the edge case you found, right?

I updated PR description to better understand fixes in this PR.

I created similar of your test first(https://github.com/aquasecurity/go-dep-parser/pull/285/commits/f7ec70961debad478ed6479ec9f16c77bb12ba79). But after investigation more i found second problem (overwrite props).

Your test only checks first fix (nested `depManagements`).
But if you revert [these changes](https://github.com/aquasecurity/go-dep-parser/blob/226facd889b9e0f9ab1debe66ec11e74ad462c66/pkg/utils/utils.go#L61-L65) (shadow overwrite props) you still pass test.

> BTW, I wanted to push the commits for testing to my fork, but pushed them to your form by mistake. Sorry 😔

No problem. I will revert these changes when we are done discussing these fixes.

knqyf263:
Thanks. I now understand what it was testing. As per your explanation, I've added properties causing the overriding bug. Could you take a look?

DmitriyLewen:
I was so focused on example from https://github.com/aquasecurity/trivy/issues/5899 that I didn't even think to just declare new property 😄

This is what i wanted to check.
Thank you for making this test easier!

knqyf263 (approved):

knqyf263:
>I was so focused on example from https://github.com/aquasecurity/trivy/issues/5899 that I didn't even think to just declare new property 😄

I wasn't part of that discussion, so I was able to focus on the essential issue 👍

PR #287, opened 2024-01-23
feat(pom): add support of line numbers for deps from base pom.xml

DmitriyLewen:
## Description
Add support of line numbers for deps from base pom.xml.

This PR doesn't add line numbers for:
- dependencies from parents
- dependencies for modules

## Related issues
- aquasecurity/trivy/issues/5945

### DIFF
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -382,10 +382,7 @@ func (p *parser) parseDependencies(deps []pomDependency, props map[string]string
 			continue
 		}
 
-		art := d.ToArtifact(opts.exclusions)
-		if !opts.lineNumber {
-			art.Locations = nil
-		}
+		art := d.ToArtifact(opts)
 
 		dependencies = append(dependencies, art)
 	}
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -382,9 +382,7 @@ func (p *parser) parseDependencies(deps []pomDependency, props map[string]string
 			continue
 		}
 
-		art := d.ToArtifact(opts)
-
-		dependencies = append(dependencies, art)
+		dependencies = append(dependencies, d.ToArtifact(opts))
 	}
 	return dependencies
 }
--- a/pkg/java/pom/parse.go
+++ b/pkg/java/pom/parse.go
@@ -92,7 +92,7 @@ func (p *parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
 	}
 
 	// Analyze root POM
-	result, err := p.analyze(root, analysisOptions{})
+	result, err := p.analyze(root, analysisOptions{lineNumber: true})
 	if err != nil {
 		return nil, nil, xerrors.Errorf("analyze error (%s): %w", p.rootPath, err)
 	}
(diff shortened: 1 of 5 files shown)


pkg/java/pom/parse.go
```diff
@@ -409,7 +433,7 @@ func (p *parser) mergeDependencies(parent, child []artifact, exclusions map[stri
 	var deps []artifact
 	unique := map[string]struct{}{}
 
-	for _, d := range append(parent, child...) {
+	for _, d := range append(child, parent...) {
```
DmitriyLewen, line 436: https://github.com/aquasecurity/go-dep-parser/blob/c95688d311ba60dfada97294d7461f07b9602521/pkg/java/pom/parse.go#L325-L328

Looks like it was a mistake.
I've updated test to show this:
https://github.com/aquasecurity/go-dep-parser/pull/287/files#diff-54393967b4dbd54d8dab7c5bf81ff4297c6fb825ff2bbf1805a4fdabec231146R44

knqyf263 (commented):
How about this approach?
https://go.dev/play/p/YYQZyFgD9Vt?v=

DmitriyLewen:
It works!
Thanks for help!

pkg/java/pom/pom.go
```diff
@@ -296,6 +308,38 @@ func (props *properties) UnmarshalXML(d *xml.Decoder, start xml.StartElement) er
 	return nil
 }
 
+func (deps *pomDependencies) UnmarshalXML(d *xml.Decoder, _ xml.StartElement) error {
+	*deps = pomDependencies{}
```
knqyf263, line 312: IIUC, we don't need this line.

DmitriyLewen: Right. Thanks! 
Changed in https://github.com/aquasecurity/go-dep-parser/pull/287/commits/522089e086271f85a67ca5d9dd61621d932eeff4

knqyf263:
@DmitriyLewen I refactored a bit so that we don't need to set nil to `Locations` in multiple places. Please review the change.
https://github.com/aquasecurity/go-dep-parser/pull/287/commits/ca6bbaa99921765f96253edb23aa69ecca19f76b

pkg/java/pom/parse.go
```diff
@@ -371,7 +381,13 @@ func (p *parser) parseDependencies(deps []pomDependency, props map[string]string
 		if (d.Scope != "" && d.Scope != "compile") || d.Optional {
 			continue
 		}
-		dependencies = append(dependencies, d.ToArtifact(exclusions))
+
+		art := d.ToArtifact(opts.exclusions)
+		if !opts.lineNumber {
+			art.Locations = nil
+		}
```
DmitriyLewen, line 388: @knqyf263  what if we move this logic to `ToArtifact`.
we can add a location if needed instead of removing it.
something like this:
```go
func (d pomDependency) ToArtifact(opts analysisOptions) artifact {
	// To avoid shadow adding exclusions to top pom's,
	// we need to initialize a new map for each new artifact
	// See `exclusions in child` test for more information
	exclusions := map[string]struct{}{}
	if opts.exclusions != nil {
		exclusions = maps.Clone(opts.exclusions)
	}
	for _, e := range d.Exclusions.Exclusion {
		exclusions[fmt.Sprintf("%s:%s", e.GroupID, e.ArtifactID)] = struct{}{}
	}

	var locs types.Locations
	if opts.lineNumber {
		locs = types.Locations{
			{
				StartLine: d.StartLine,
				EndLine:   d.EndLine,
			},
		}
	}
	return artifact{
		GroupID:    d.GroupID,
		ArtifactID: d.ArtifactID,
		Version:    newVersion(d.Version),
		Exclusions: exclusions,
		Locations:  locs,
	}
}
```

knqyf263: Yes, I also had the idea in my mind. Both work for me. You can decide it.

DmitriyLewen: Added changes in https://github.com/aquasecurity/go-dep-parser/pull/287/commits/e8ff04cc7449918aa92adce18250fa3455cb8f18

DmitriyLewen:
It looks good to me now.

knqyf263 (approved):

Issue #286, opened 2024-01-22
Support for pom file parsing for maven repos with username/password


renypaul:
Currently maven repos with credentials doesn't seems to be supported for fetching the packages. Adding the support will be useful for private repositories like nexus

PR #289, opened 2024-01-31
Add support for fetching from repository mentioned in pom file

renypaul:
Issue: https://github.com/aquasecurity/go-dep-parser/issues/286
  ## Description
  This change allows packages to be fetched from repository other than https://repo.maven.apache.org/maven2/
  - Get username and password from settings file for the corresponding server ID mentioned in the pom file
  - Added support for authentication if username and password mentioned while fetching the package


pkg/java/pom/settings.go
```diff
@@ -8,24 +8,43 @@ import (
...
-	if err == nil && globalSettings.LocalRepository != "" {
-		return globalSettings
+	if err == nil {
+		if globalSettings.LocalRepository != "" {
+			return globalSettings
+		}
+		if !settingsFound {
+			s = globalSettings
+		}
 	}
```
DmitriyLewen, line 45: Looks like we need to merge global and user settings - `The former settings.xml are also called global settings, the latter settings.xml are referred to as user settings. If both files exists, their contents gets merged, with the user-specific settings.xml being dominant.` (https://maven.apache.org/settings.html#quick-overview)
order for `LocalRepository` : `user` -> `global`
But for repositories we should use `global` -> `user` order - https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order

pkg/java/pom/parse.go
```diff
@@ -52,6 +52,7 @@ type parser struct {
 	localRepository    string
 	remoteRepositories []string
 	offline            bool
+	settings           settings
```
DmitriyLewen, line 55: we duplicate `localRepository`.
It might be better to just use serverSettings(`[]Server`).

pkg/java/pom/parse.go
```diff
@@ -86,6 +88,24 @@ func (p *parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+		for _, server := range p.settings.Servers {
+			if rep.ID == server.ID && server.Username != "" && server.Password != "" {
+				repoURL.User = url.UserPassword(server.Username, server.Password)
+				break
+			}
+		}
+
+		if rep.Releases.Enabled != "false" {
+			p.remoteRepositories = append(p.remoteRepositories, repoURL.String())
+		}
```
DmitriyLewen, line 106: It looks like we can't use repositories only for enabled `releases` - https://maven.apache.org/settings.html#repositories.
IIUC we need to check `pluginRepositories` along with `releases`.

But this looks like rare case. Perhaps we can start by adding all repositories.

DmitriyLewen (commented):
Hello @renypaul 
Thanks for your work!

I left some comments.
Can you also added some tests for these changes?

Regards, Dmitriy

pkg/java/pom/settings.go
renypaul: Thanks @DmitriyLewen for the review and pointers. I have updated the code based on your suggestion. Please take a look

pkg/java/pom/parse.go
renypaul: Added pluginRepositories also

DmitriyLewen: hm... i re-read maven docs and it seems I made mistake and confused you.
`pluginRepositories` is used only for set up repositories to downloading plugins.
 But we don't check plugins, so we don't need this.

But I thought about “snapshots.” What if we add a repository if `release` or `snapshot` is enabled:
```go
for _, rep := range content.Repositories.Repository {
		if rep.Releases.Enabled == "false" || rep.Snapshots.Enabled == "false" {
			continue
		}

		repoURL, err := url.Parse(rep.URL)
		if err != nil {
			continue
		}

		for _, server := range p.servers {
			if rep.ID == server.ID && server.Username != "" && server.Password != "" {
				repoURL.User = url.UserPassword(server.Username, server.Password)
				break
			}
		}

		log.Logger.Debugf("Adding repository %s: %s", rep.ID, rep.URL)
		p.remoteRepositories = append(p.remoteRepositories, repoURL.String())
	}
```

pkg/java/pom/settings.go
DmitriyLewen: These changes seem good to me.
Some moments:
1 - Add comments with links to order, please.
2. I'm not sure if `/usr/share/maven` is default path. I didn't find any information on this path. Do you have docs on this matter?
If I remember correctly, you set MAVEN_HOME when you install maven. So MAVEN_HOME must exist.

renypaul: - Added comments and links to document
- Added testcase
- What I found is that MAVEN_HOME is not not mandatory and not always set. 
- `mvn` find the home based on how/where its installed unless MAVEN_HOME variable is set. 
- `/usr/share/maven` is not always the default path and it depends on how maven in installed. For Ubuntu installation with apt it is `/usr/share/maven` , but if we download the tar.gz from maven and install, it will be the extracted folder. 
```/usr/bin/mvn -V
Apache Maven 3.6.3
Maven home: /usr/share/maven

./mvn -V
Apache Maven 3.9.6 (bc0240f3c744dd6b6ec2920b3cd08dcc295161ae)
Maven home: /home/reny/apache-maven-3.9.6```
So if we keep the default value as `/usr/share/maven` it will get the settings in some OS where if its not there, the settings will not get it anywhere. Let me know if you need the default value to be removed @DmitriyLewen

File: pkg/java/pom/parse.go [hunk 4]
```diff
@@ -86,6 +88,32 @@ func (p *parser) Parse(r dio.ReadSeekerAt) ([]types.Library, []types.Dependency,
...
+				break
+			}
+		}
+
+		log.Logger.Debugf("Adding repository %s: %s", rep.ID, rep.URL)
+		remoteRepositories = append(remoteRepositories, repoURL.String())
+	}
+
+	// Add central maven repository or repositories obtained using `WithRemoteRepos` function.
+	remoteRepositories = append(remoteRepositories, p.remoteRepositories...)
```
DmitriyLewen, line 115: I'm not sure about this.
Do we need to always add central Maven repository or only when remote repositories are not used.
@renypaul wdyt?

DmitriyLewen (commented):
I refactored a little.
@renypaul Can you take a look please?

pkg/java/pom/parse.go
renypaul: @DmitriyLewen If we don't mention maven default repository, mvn command still pulls from maven repository. So I believe we should keep the behavior similar to mvn and keep default maven repository

renypaul: @DmitriyLewen Thanks for refactoring the code. It looks like the expected result is slightly different. I will check and update the test code if needed

DmitriyLewen: > So I believe we should keep the behavior similar to mvn and keep default maven repository

But if remote repositories are used, then `maven` will check the artifact from those repositories first, right?

DmitriyLewen: > update the test code if needed

Thanks!

renypaul: ### Following is the output from `mvn help:effective-settings`

This is based on the command mentioned in https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order
```
root@go~(PASS)$ mvn help:effective-settings
[INFO] ------------------< org.apache.maven:standalone-pom >-------------------
[INFO] --- maven-help-plugin:3.4.0:effective-settings (default-cli) @ standalone-pom ---
Effective user-specific configuration settings:

<?xml version="1.0" encoding="UTF-8"?>
<settings xmlns="http://maven.apache.org/SETTINGS/1.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd">
  <localRepository>/root/testdata/user/repository</localRepository>
  <servers>
    <server>
      <id>user-server</id>
    </server>
    <server>
      <username>test-user</username>
      <password>***</password>
      <id>server-with-credentials</id>
    </server>
    <server>
      <username>test-user-only</username>
      <id>server-with-name-only</id>
    </server>
    <server>
      <id>global-server</id>
    </server>
  </servers>
  <pluginGroups>
    <pluginGroup>org.apache.maven.plugins</pluginGroup>
    <pluginGroup>org.codehaus.mojo</pluginGroup>
  </pluginGroups>
</settings>

root@go~(PASS)$ cat ~/.m2/settings.xml
<?xml version="1.0" encoding="UTF-8"?>
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
    <localRepository>testdata/user/repository</localRepository>

    <servers>
        <server>
            <id>user-server</id>
        </server>
        <server>
            <id>server-with-credentials</id>
            <username>test-user</username>
            <password>test-password</password>
        </server>
        <server>
            <id>server-with-name-only</id>
            <username>test-user-only</username>
        </server>
    </servers>
</settings>
root@go~(PASS)$ cat /usr/share/maven/conf/settings.xml
<?xml version="1.0" encoding="UTF-8"?>
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
    <localRepository>testdata/repository</localRepository>

    <servers>
        <server>
            <id>global-server</id>
        </server>
        <server>
            <id>server-with-credentials</id>
            <username>test-user-from-global</username>
            <password>test-password-from-global</password>
        </server>
        <server>
            <id>server-with-name-only</id>
            <username>test-user-only</username>
        </server>
    </servers>
</settings>
```
From the output, the summary is
- mvn looks for server-id only while checking for repository and takes rest of the configuration from the settings.xml where it found first
- mvn give preference to user settings for server id also
@DmitriyLewen Based on the result, I am thinking of updating the code. Please let me know if I am missing something

DmitriyLewen: > Based on the result, I am thinking of updating the code. Please let me know if I am missing something

hm... it looks like there are mistakes in the documentation...
User servers > global servers.
It will be great if you update the code.
Can you also add the output of the `mvn help:effective-settings` command to test case (e.g. `hard requirement for the specified version` test) and remove link to docs about servers (to avoid confusion)?

pkg/java/pom/settings_test.go
```diff
@@ -0,0 +1,111 @@
...
+					},
+					{
+						ID:       "server-with-name-only",
+						Username: "test-user-only",
+					},
+				},
+			},
+		},
+		{
+			// Expected result is the output of mvn help:effective-settings
```
DmitriyLewen, line 65: To understand order, you do not need to install `mvn` and run `mvn help:effective-settings`.
```suggestion
			// $ mvn help:effective-settings
			//[INFO] ------------------< org.apache.maven:standalone-pom >-------------------
			//[INFO] --- maven-help-plugin:3.4.0:effective-settings (default-cli) @ standalone-pom ---
			//Effective user-specific configuration settings:
			//
			//<?xml version="1.0" encoding="UTF-8"?>
			//<settings xmlns="http://maven.apache.org/SETTINGS/1.1.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd">
			//  <localRepository>/root/testdata/user/repository</localRepository>
			//  <servers>
			//    <server>
			//      <id>user-server</id>
			//    </server>
			//    <server>
			//      <username>test-user</username>
			//      <password>***</password>
			//      <id>server-with-credentials</id>
			//    </server>
			//    <server>
			//      <username>test-user-only</username>
			//      <id>server-with-name-only</id>
			//    </server>
			//    <server>
			//      <id>global-server</id>
			//    </server>
			//  </servers>
			//</settings>
```

pkg/java/pom/settings.go
```diff
@@ -8,24 +8,58 @@ import (
...
+	Password string `xml:"password"`
+}
+
 type settings struct {
-	LocalRepository string `xml:"localRepository"`
+	LocalRepository string   `xml:"localRepository"`
+	Servers         []Server `xml:"servers>server"`
+}
+
+func serverFound(servers []Server, id string) bool {
```
DmitriyLewen, line 22: let's write about Maven comparison method.
```suggestion
// serverFound checks that servers already contain server.
// Maven compares servers by ID only.
func serverFound(servers []Server, id string) bool {
```

pkg/java/pom/settings.go
```diff
@@ -8,24 +8,58 @@ import (
...
-	if err == nil && globalSettings.LocalRepository != "" {
-		return globalSettings
+	if err == nil {
+		// We need to merge global and user settings. User settings being dominant.
+		// https://maven.apache.org/settings.html#quick-overview
+		if s.LocalRepository == "" {
+			s.LocalRepository = globalSettings.LocalRepository
+		}
+		// Servers are added based on output of `mvn help:effective-settings`.
+		// https://maven.apache.org/guides/mini/guide-multiple-repositories.html#repository-order
```
DmitriyLewen, line 54: Docs are wrong. Let's just write about the order.
```suggestion
		// Maven checks user servers first, then global servers
```

pkg/java/pom/parse.go
DmitriyLewen: Thanks for your work!
I left some nil comments. Take a look, please.

DmitriyLewen (approved):
@renypaul Thanks a lot for your work!
I approved this PR.
@knqyf263 take a look and merge this, when you have time, please.

renypaul:
Thanks @DmitriyLewen for the careful review and suggestions

renypaul:
@knqyf263 Gentle reminder on merge. Also it will be helpful to know when we can expect trivy with this change. FYI @DmitriyLewen

knqyf263:
Hopefully, I'll check it next week.

renypaul:
> Hopefully, I'll check it next week.

Gentle reminder @knqyf263

DmitriyLewen:
Hello @renypaul 

We have moved `go-dep-parser` to Trivy - https://github.com/aquasecurity/trivy/tree/main/pkg/dependent/parser.

Do you have time to open a new PR for Trivy with these changes?
If not, tell me, I will do it.

renypaul:
> Hello @renypaul
> 
> We have moved `go-dep-parser` to Trivy - https://github.com/aquasecurity/trivy/tree/main/pkg/dependent/parser.
> 
> Do you have time to open a new PR for Trivy with these changes? If not, tell me, I will do it.

Thanks @DmitriyLewen  I have created a PR at https://github.com/aquasecurity/trivy/pull/6171/files . It will be great if this can be included in the next release. Please feel free to update the code if needed.

DmitriyLewen:
`go-dep-parser` has been moved to Trivy.
https://github.com/aquasecurity/trivy/pull/6171 was created for these changes.

